PR #23, opened 2021-08-30
Add TypeScript plugin

### LINKED ISSUE #21: Allow Types in JavaScript/TypeScript files
### Background & Motivation

When creating a project that reexports Astro components, it would be nice if the types followinged, for example:

__Name.astro__

```astro
---
export interface Props {
  name: string;
}
---
```

__index.ts__

```ts
import Name from './Name.astro';

export {
  Name
};
```

However, currently the props do not follow through into TypeScript. 

### Proposed Solution

Build a TypeScript plugin which allows TypeScript files to consume Astro types.

### Detailed Design

_No response_

### Help make it happen!

- [x] I am willing to submit a PR to implement this change.
- [ ] I am willing to submit a PR to implement this change, but would need some guidance.
- [ ] I am not willing to submit a PR to implement this change.

matthewp:
## Changes

This adds a new package, `@astrojs/ts-plugin` which is a TypeScript plugin. It runs whenever a ts/tsx/js/jsx file is open in VSCode and provides type information for .astro files, allowing prop completion from .astro files.

Loom:

https://www.loom.com/share/68ec7de42cee4fc1b20083012a2c2c57

Closes #21

### DIFF
--- a/packages/ts-plugin/src/astro-sys.ts
+++ b/packages/ts-plugin/src/astro-sys.ts
@@ -1,0 +1,32 @@
+import ts from 'typescript';
+import { Logger } from './logger.js';
+import { ensureRealAstroFilePath, isVirtualAstroFilePath, toRealAstroFilePath } from './utils.js';
+
+/**
+ * This should only be accessed by TS svelte module resolution.
+ */
+export function createAstroSys(logger: Logger) {
+    const astroSys: ts.System = {
+        ...ts.sys,
+        fileExists(path: string) {
+            return ts.sys.fileExists(ensureRealAstroFilePath(path));
+        },
+        readDirectory(path, extensions, exclude, include, depth) {
+            const extensionsWithAstro = (extensions ?? []).concat('.astro');
+
+            return ts.sys.readDirectory(path, extensionsWithAstro, exclude, include, depth);
+        }
+    };
+
+    if (ts.sys.realpath) {
+        const realpath = ts.sys.realpath;
+        astroSys.realpath = function (path) {
+            if (isVirtualAstroFilePath(path)) {
+                return realpath(toRealAstroFilePath(path)) + '.ts';
+            }
+            return realpath(path);
+        };
+    }
+
+    return astroSys;
+}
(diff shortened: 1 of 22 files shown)


tony-sull (changes requested):
Nice, lgtm!  Flagged a few comments I noticed referencing Svelte instead of Astro, but all the actual code looks great.

Can't wait to finally put all the `Props` interfaces I've been writing to good use :laughing: :

packages/ts-plugin/src/astro2tsx.ts
```diff
@@ -0,0 +1,49 @@
...
+  };
+  
+  return result;
+}
+
+// This is hacky but it works for now
+function addProps(content: string, dtsContent: string): string {
+  let defaultExportType = 'Record<string, any>';
+  if(/interface Props/.test(content)) {
+    defaultExportType = 'Props';
```
FredKSchott, line 38: blah, yea this is pretty gross. I imagine our new parser should eventually be able to give us this information, so that you could just do something like `const {ast} = awair parse(code)` and have the exported types included.

FredKSchott: agreed though this is fine for now, as a first pass

FredKSchott: another short-term option: https://stackoverflow.com/a/38086176/4682119

matthewp: The new compiler doesn't parse the component script section so it's not going to give this info, will still need to run it through the ts compiler API. I can probably add this now though, to cover cases where it's not an interface for example.

packages/ts-plugin/src/astro-sys.ts
```diff
@@ -0,0 +1,32 @@
+import ts from 'typescript';
+import { Logger } from './logger.js';
+import { ensureRealAstroFilePath, isVirtualAstroFilePath, toRealAstroFilePath } from './utils.js';
+
+/**
+ * This should only be accessed by TS svelte module resolution.
```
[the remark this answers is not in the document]
matthewp: my find/replace skills are weak, I'll go over this all again 😀

packages/ts-plugin/src/astro2tsx.ts
matthewp: Ok, brought in ts-morph and am checking for the Props export based on the AST now, works with interfaces or types.

PR #32, opened 2021-09-02
Improve perf for all operations

### LINKED ISSUE #31: 🐛 BUG: Performance
### Describe the Bug

Autocompletion performance is often slow

https://github.com/sveltejs/language-tools/pull/1022/files

### Steps to Reproduce

1. `npm init astro` using template <NAME>
2. ...
3. ...
4. ...
5. Error! Describe what went wrong (and what was expected instead)...

matthewp:
## Changes

- TypeScript internally calls us to a "sys" module that we own to search for files. We were telling it that any non-astro file needed to have a .ts extension appended. So it would keep searching for files, appending .ts many times, until ultimately giving up, but this took 300+ms.
- This fixes it by only appending the .ts for our virtual astro ts modules.
- This brings the onCompletion time down to 1-30ms.

Closes #31 

## Testing

N/A

## Docs

N/A

### DIFF
--- a/packages/language-server/src/plugins/typescript/utils.ts
+++ b/packages/language-server/src/plugins/typescript/utils.ts
@@ -178,10 +178,13 @@ export function isVirtualFilePath(filePath: string) {
 }
 
 export function toVirtualAstroFilePath(filePath: string) {
-  if (isVirtualFrameworkFilePath('astro', filePath)) {
+  if (isVirtualAstroFilePath(filePath)) {
+    return filePath;
+  } else if(isAstroFilePath(filePath)) {
+    return `${filePath}.ts`;
+  } else {
     return filePath;
   }
-  return `${filePath}.ts`;
 }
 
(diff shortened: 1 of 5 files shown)


packages/language-server/src/plugins/typescript/utils.ts
```diff
@@ -178,10 +178,13 @@ export function isVirtualFilePath(filePath: string) {
 }
 
 export function toVirtualAstroFilePath(filePath: string) {
-  if (isVirtualFrameworkFilePath('astro', filePath)) {
+  if (isVirtualAstroFilePath(filePath)) {
+    return filePath;
+  } else if(isAstroFilePath(filePath)) {
```
matthewp, line 183: This is what fixes perf, this else if was missing before and causing TS to search for files for a long time.

FredKSchott:
LGTM! Ship it!

FredKSchott (approved):

PR #39, opened 2021-09-09
Provide different highlighting for component names

### LINKED ISSUE #4: VSCode: Components should be highlighted with a different color
### What package manager are you using?

yarn

### What operating system are you using?

Windows

### Describe the Bug

Components are highlighted with the same color as general HTML codes.

### Steps to Reproduce

Components should be highlighted with a different color.

### Link to Minimal Reproducible Example (Optional)

_No response_

matthewp:
## Changes

- Gives component names a slightly different highlighting style, mirroring what Svelte does
- Closes #4

Example of new component highlighting

### DIFF
--- a/.changeset/real-parrots-ring.md
+++ b/.changeset/real-parrots-ring.md
@@ -1,0 +1,5 @@
+---
+"astro-vscode": patch
+---
+
+Provides special highlighting for component names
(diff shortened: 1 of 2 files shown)


aFuzzyBear:
This neurotic pain in the ass tool, is a good idea of playing about with colours in a vs code theme, Ive got a few as a WIP, which ive yet to get back to, but its handy to get something started, definitely not a finessing tool, 
https://themes.vscode.one/

matthewp:
Just to clarify, this is how it looks in *my* theme, it will look differently depending on what you are using, but is using the same token as what Svelte uses for its vscode extension, so I think this is a sane way to go.

PR #101, opened 2021-11-04
fix: ignore tsconfig setting files

### LINKED ISSUE #91: 🐛 BUG: `strict` tsconfig rule causes linting errors in `.astro` files
### What package manager are you using?

yarn

### What operating system are you using?

macOS

### Describe the Bug

When using a `tsconfig.json` file such as :

```
{
   "compilerOptions": {
      "strict": true,
      "alwaysStrict": true
   }
}
```

Visual Studio Code highlights all of the JSX and shows this linting error:

`JSX element implicitly has type 'any' because no interface 'JSX.IntrinsicElements' exists.`

image

In Discord, Matthew2 suspects this might be a bug with the Astro extension.

https://discord.com/channels/830184174198718474/845451724738265138/895704312544653353

### Steps to Reproduce

1. `npm init astro --template minimal` using template <NAME>
2. Open `index.astro` and note the lack of linting warnings.
3. Add the `tsconfig.json` sample above.
4. Close and then reopne `index.astro`.
5. Notice all the linting errors.

### Link to Minimal Reproducible Example (Optional)

https://github.com/justinnoel/astro-jsx-linting-issue

justinnoel:
Astro should not be reading the tsconf settings file.  This causes issues  such as `strict: true` causing ts errors like:

`JSX element implicitly has type 'any' because no interface 'JSX.IntrinsicElements' exists.`

RE: https://discord.com/channels/830184174198718474/845451724738265138/895704312544653353

Closes #91.

## Changes

- Stops using the `tsconfig.json` as part of the default configuration for the Astro language tools

## Testing

- Tested manually be using `strict: true` in the `tsconfig.json` file before changes.
  - The Visual Studio Code editor highlighted all the JSX code with the  `JSX element implicitly has type 'any' because no interface 'JSX.IntrinsicElements' exists.` error
- After changes, Code did not produce the error.
- Also manually tested creating constants, using Astro props etc to ensure TypeScript still worked properly in `.astro` files.  

## Docs

- No doc changes are required.

cc: @matthewp

### DIFF
--- a/.changeset/fall-leaves-fall.md
+++ b/.changeset/fall-leaves-fall.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/language-server": minor
+---
+
+Ignore tsconfig setting files
--- a/packages/language-server/src/plugins/typescript/languageService.ts
+++ b/packages/language-server/src/plugins/typescript/languageService.ts
@@ -55,18 +55,7 @@ async function createLanguageService(tsconfigPath: string, workspaceRoot: string
     },
   };
 
-  let configJson = (tsconfigPath && ts.readConfigFile(tsconfigPath, ts.sys.readFile).config) || getDefaultJsConfig();
-  if (!configJson.extends) {
-    configJson = Object.assign(
-      {
-        exclude: getDefaultExclude(),
-      },
-      configJson
-    );
-  }
-
-  // Delete include so that astro files don't get excluded.
-  delete configJson.include;
+  const configJson = getDefaultJsConfig();
 
   const existingCompilerOptions: ts.CompilerOptions = {
     jsx: ts.JsxEmit.Preserve,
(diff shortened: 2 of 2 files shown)


matthewp:
Looks great! Context here is that Astro files have caused people with different tsconfig settings quite a bit of pain because those config settings are meant to apply to .ts/.tsx files and by applying them to `.astro` files it causes false-positive errors.

@justinnoel can you add a changeset by running `yarn changeset` and going through the prompts? Thanks!

matthewp:
Hm, not reading some of what people are doing in discord makes me think maybe we *should* still be reading their config: https://discord.com/channels/830184174198718474/871832088263933973/905904428538789888

Can you try removing your change and instead overriding `strict`? Would doing that fix your issues?

justinnoel:
> Hm, not reading some of what people are doing in discord makes me think maybe we _should_ still be reading their config: https://discord.com/channels/830184174198718474/871832088263933973/905904428538789888
> 
> Can you try removing your change and instead overriding `strict`? Would doing that fix your issues?

Looking at this now. Definitely seems like we need to honor `tsconfig` and this fix is too nuclear.

justinnoel:
Pretty sure I found a different solution to this whole JSX problem. 

I'm going to park this PR for now and submit a different one. If the new one is the right solution, I'll come back and close this one.

justinnoel:
I'm closing this now because I've found a way to resolve the issues with `strict: true` implementation.

I've setup a sample repo that converts the standard Astro Blog template to use `strict: true`:

https://github.com/justinnoel/astro-blog-strict

PR #113, opened 2021-11-24
Support template literal attribute values

jonathantneal:
## Changes

- Adds support for template literal attribute values. 😄
- Adds more fragility to the already-fragile RegExp patches. 😅

```astro
---
import MyComponent from '../components/MyComponent.astro'
---
<MyComponent mine=`yes` />

## Testing

no tests added

## Docs

bug fix only


jonathantneal:
Closing. This would require other changes to be done right.

PR #133, opened 2021-12-13
Add Node v12 support

FredKSchott:
## Changes

- Updates tests to run on Node v12
- Updates tsconfig to support node v12
- Updates CI to also test Node v12
- Fixes the new Astro smoke test: https://github.com/withastro/astro/runs/4480411520?check_suite_focus=true

## Testing

- tested!

## Docs

- N/A

### DIFF
--- a/.changeset/soft-pens-glow.md
+++ b/.changeset/soft-pens-glow.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/language-server": patch
+---
+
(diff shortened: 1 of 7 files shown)


matthewp:
lgtm. Failure is because of the tests using top-level await. Can fix by wrapping in a IIFE:

```js
(async function() {
  // Tests here
})();
```

PR #145, opened 2022-01-04
Remove import.meta declarations from our d.ts

matthewp:
## Changes

- Removes declarations overriding `import.meta.hot` and `import.meta.env` because these declarations are already provided by Vite.
- This removes errors that you see in the Astro repo.

## Testing

Not testable.

## Docs

N/A


jonathantneal (approved):
Clean.

PR #148, opened 2022-01-10
Update project from astro-language-tools to language-tools

jonathantneal:
## Changes

- Updates any relevant assets to point to https://github.com/withastro/language-tools
- Conforms `language-tools` with `compiler` (previously `astro-compiler`).

## Testing

name update only

## Docs

name update only


jonathantneal:
Chatted with @matthewp and agreed today is the day to move the GitHub URL.

PR #150, opened 2022-01-13
Improve import assertion support

jonathantneal:
## Changes

- Improves `assert` syntax highlighting
- Aligns syntax highlighting with pre-existing capabilities

## Testing

**Before**:
before syntax highlighting fix
- Keyword `assert` was displayed as a variable.
- Assertion object was displayed as a series of variables and delimiters.

**After**:
after syntax highlighting fix

- Keyword `assert` is displayed as a keyword. 
- Assertion object is displayed as an object with a key and a string member.

### DIFF
--- a/.changeset/itchy-chefs-cover.md
+++ b/.changeset/itchy-chefs-cover.md
@@ -1,0 +1,6 @@
+---
+"@astrojs/language-server": patch
+"astro-vscode": patch
+---
+
+Improve import assertion syntax highlighting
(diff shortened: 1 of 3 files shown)


FredKSchott:
Chatted in person today, but unless we know how we want to use this as a project and have plans to merge support for this soon, can we hold off on this? Implementation wise it seems to be injecting behavior on top of the default JS syntax, and I worry what will happen when JS syntax highlighting adds its own support for this. Would love to keep complexity down if we can help it.

jonathantneal:
We can certainly hold off on it, and either allow VSCode to do what we have done in some fashion, or move this code into a PR to [microsoft/vscode](https://github.com/microsoft/vscode). This was just an isolated highlighting fix.

PR #151, opened 2022-01-15
Add `types: ['vite/client']` to the default tsconfig used for .astro files

Princesseuh:
## Changes

This adds `types: ['vite/client']` to the default TypeScript config the extension uses for .astro files. What this provide us with is nice typing for `import.meta.env` without the user needing to do anything on their end

Additionally, the user can provide their own typing through [the method Vite recommend](https://vitejs.dev/guide/env-and-mode.html#intellisense-for-typescript), it work just fine (as far as I tested), as demonstrated by the screenshot below

image

### Caveats

This only add the typing for `.astro` files, the user still need to provide a `tsconfig.json` with `types: ['vite/client']` for support in other file formats (notably in our case, that includes .ts, .vue, .svelte etc files). As far as I know, they do not need to add `vite` as a dependency of their project, just the types in their tsconfig

On a personal level, I do not think this is a problem as long as it's properly documented somewhere. I tried to search for a way to fix this but couldn't find anything, if someone has an idea, I'm open to adding it to this PR (might need a bit of guidance however)

## Testing

Tested manually using the `minimal` example with all three majors package manager (npm, yarn, pnpm)

## Docs

No docs needed, I added a comment to explain why it's been added there however (as it's a bit out of line with the other default settings)

### DIFF
--- a/packages/language-server/src/plugins/typescript/languageService.ts
+++ b/packages/language-server/src/plugins/typescript/languageService.ts
@@ -171,6 +171,8 @@ function getDefaultJsConfig(): {
       maxNodeModuleJsDepth: 2,
       allowSyntheticDefaultImports: true,
       allowJs: true,
+      // By providing vite/client here, our users get proper typing on import.meta in .astro files
+      types: ['vite/client']
    };
    return {
       compilerOptions,


jonathantneal (approved):

jonathantneal:
This seems like a good add, @Princesseuh. Thanks for the notes about its current functionality and limitations. Once the tests run (and pass, presumably), I’ll merge.

quantuminformation:
do we need update to package json?

Princesseuh:
> do we need update to package json? this didnt help me

Hello, this issue is fairly old and things don't really work the same way anymore. 

What issues are you encountering? `import.meta.env` should be typed by default in any Astro projects nowadays

quantuminformation:
im using solid, let me do a diff on the latest 

my issue:

image

Princesseuh:
Ah alright, so this issue (and this repo in general) is only about `.astro` files. 

For your issue, here's what to do:
- Make sure your project includes the proper types, either through a `.d.ts` inside your project ([example](https://github.com/withastro/astro/blob/main/examples/basics/src/env.d.ts)) or through your `tsconfig.json`'s `types`. [See the part below the config template here](https://docs.astro.build/en/guides/typescript/#setup) for more information.
- If your project does includes the proper types and it still doesn't work, make sure to be using at least version `1.1.1` of Astro, as there was an issue with getting those types before that version

Don't hesitate if you have any further questions / issues!

PR #144, opened 2022-01-04
fix: ignore tsconfig setting files

matthewp:
Astro should not be reading the tsconf settings file.  This causes issues  such as `strict: true` causing ts errors like:

`JSX element implicitly has type 'any' because no interface 'JSX.IntrinsicElements' exists.`

RE: https://discord.com/channels/830184174198718474/845451724738265138/895704312544653353

Closes #91.

Personal References:
- Link #T-366
- Closes #T-366

PR #157, opened 2022-01-20
Bring back loading the user js/tsconfig.json

### LINKED ISSUE #156: 🐛 BUG: Aliases not working
### Describe the Bug

Aliases do not get any IntelliSense and throw an error when I try to "Go to Definition". However the docs state:
> These aliases are also integrated automatically into VSCode and other editors.

So am I doing something wrong? In `astro dev` everything seems to work but not in the editor.

### Steps to Reproduce

1. `npm init astro` with Minimal template
2. `npm i`
3. Make `tsconfig.json` look like this:
```json
{
  "compilerOptions": {
    "moduleResolution": "node",
    "baseUrl": ".",
    "paths": {
      "component:*": ["src/components/*.astro"]
    }
  }
}
```
4. Create file `src/components/Example.astro`
5. Make `src/pages/index.astro` look like this:
```astro
---
import Example from "component:Example";
---

<html lang="en">
<head>
    <meta charset="utf-8" />
    <meta name="viewport" content="width=device-width" />
    <title>Welcome to Astro</title>
</head>
<body>
</body>
</html>
```
6. When you right click > "Go to Definition" on the variable `Example` in `index.astro` VSCode throws an error:
> Unable to open 'component:Example': Unable to read file 'vscode-remote://wsl+ubuntu/home/igor/d/astro-alias-report/src/pages/component:Example' (Error: Unable to resolve nonexistent file 'vscode-remote://wsl+ubuntu/home/igor/d/astro-alias-report/src/pages/component:Example').

Also, after trying to "Go to Definition" now the `"component:Example"` string has red squiggles with the error message:
> File '/home/igor/d/astro-alias-report/src/pages/component:Example.ts' is not a module.

Princesseuh:
## Changes

This bring back the feature of loading the user `js/tsconfig.json` that was removed in https://github.com/withastro/language-tools/pull/144

It was removed because users with `strict: true` or `alwaysStrict: true` in their config experienced unrelated errors in .astro files (see https://github.com/withastro/language-tools/issues/91). However removing it causes other issues, most importantly, aliases don't work anymore (https://github.com/withastro/language-tools/issues/156)

This PR fix this by bringing back loading the user config but forcing `strict: false` for .astro files. Additionally, unlike the previous solution, we gently add some of our default settings through merging with the user config (so they're overwrittable if needed), for instance `type: ["vite/client"]` that was added in https://github.com/withastro/language-tools/pull/151. This way, even if the user has a .tsconfig.json file only for ex: aliases, they'll still get the benefits of having types for ImportMeta in their .astro files

This fix #156 

## Testing

Tested manually

## Docs

No docs needed, for now at least. It'd probably be great to document how this operate in some way (perhaps in the extension README that shows in VSCode?)

### DIFF
--- a/packages/language-server/src/plugins/typescript/languageService.ts
+++ b/packages/language-server/src/plugins/typescript/languageService.ts
@@ -55,12 +55,33 @@ async function createLanguageService(tsconfigPath: string, workspaceRoot: string
     },
   };
 
-  const configJson = getDefaultJsConfig();
+  let configJson = (tsconfigPath && ts.readConfigFile(tsconfigPath, ts.sys.readFile).config) || {};
 
+  // If our user has types in their config but it doesn't include the types for ImportMeta, let's add them for them
+  if (
+    configJson.compilerOptions?.types &&
+    !configJson.compilerOptions?.types.includes("vite/client")
+  ) {
+    configJson.compilerOptions.types.push("vite/client");
+  }
+
+  configJson.compilerOptions = Object.assign(
+    getDefaultCompilerOptions(),
+    configJson.compilerOptions
+  );
+  // If the user supplied exclude, let's use theirs
+  configJson.exclude ?? (configJson.exclude = getDefaultExclude());
+
+  // Delete include so that .astro files don't get mistakenly excluded by the user
+  delete configJson.include;
+
+  // Everything here will always, unconditionally, be in the resulting config, not the opposite, tricky
   const existingCompilerOptions: ts.CompilerOptions = {
+    // Setting strict to true for .astro files leads to a lot of unrelated errors (see language-tools#91) so we force it off for .astro files
+    strict: false,
     jsx: ts.JsxEmit.Preserve,
     module: ts.ModuleKind.ESNext,
-    target: ts.ScriptTarget.ESNext
+    target: ts.ScriptTarget.ESNext,
   };
 
   const project = ts.parseJsonConfigFileContent(configJson, parseConfigHost, workspaceRoot, existingCompilerOptions, basename(tsconfigPath), undefined, [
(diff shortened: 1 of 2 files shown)


packages/language-server/src/plugins/typescript/languageService.ts
```diff
@@ -55,12 +55,33 @@ async function createLanguageService(tsconfigPath: string, workspaceRoot: string
...
+  ) {
+    configJson.compilerOptions.types.push("vite/client");
+  }
+
+  configJson.compilerOptions = Object.assign(
+    getDefaultCompilerOptions(),
+    configJson.compilerOptions
+  );
+  // If the user supplied exclude, let's use theirs
+  configJson.exclude ?? (configJson.exclude = getDefaultExclude());
```
Princesseuh, line 73: I wasn't sure how to operate for exclude so I opted for this solution

TL;DR: If the user supplies `exclude` user theirs, otherwise use ours

packages/language-server/src/plugins/typescript/languageService.ts
```diff
@@ -55,12 +55,33 @@ async function createLanguageService(tsconfigPath: string, workspaceRoot: string
...
+
+  // If our user has types in their config but it doesn't include the types for ImportMeta, let's add them for them
+  if (
+    configJson.compilerOptions?.types &&
+    !configJson.compilerOptions?.types.includes("vite/client")
+  ) {
+    configJson.compilerOptions.types.push("vite/client");
+  }
+
+  configJson.compilerOptions = Object.assign(
```
Princesseuh, line 68: Using `Object.assign` mean that this is a shallow merge, I'm not sure if that could cause issues but I don't think so?

Princesseuh:
The previous implementation had this part in it:

```js
if (!configJson.extends) {
  configJson = Object.assign(
    {
      exclude: getDefaultExclude(),
    },
    configJson
  );
}
```

But I couldn't see the point and Matthew didn't remember why it was added so, I didn't add it back. I'm not quite sure if there could be a potential incompatibility with `extends` but I couldn't find any 🤷‍♀️

packages/language-server/src/plugins/typescript/languageService.ts
matthewp: Nope, should be ok

packages/language-server/src/plugins/typescript/languageService.ts
```diff
@@ -55,12 +55,33 @@ async function createLanguageService(tsconfigPath: string, workspaceRoot: string
...
+    getDefaultCompilerOptions(),
+    configJson.compilerOptions
+  );
+  // If the user supplied exclude, let's use theirs
+  configJson.exclude ?? (configJson.exclude = getDefaultExclude());
 
+  // Delete include so that .astro files don't get mistakenly excluded by the user
+  delete configJson.include;
+
+  // Everything here will always, unconditionally, be in the resulting config, not the opposite, tricky
```
matthewp, line 78: great comments

matthewp (approved):
lgtm, thanks so much!

PR #159, opened 2022-01-22
Refactor README and other .md files

Princesseuh:
## Motivations

As someone new to working on language servers and VS Code extensions, the structure of this repo was a bit confusing to me. Additionally, I felt that the current README were a bit barebones (which is fair, it's not really the most important part)

I figured that trying to document it a bit would be a good way to learn my way around and it did help a lot actually. After this PR I feel much more ready to tackle bigger tasks

(Also, writing issues, descriptions, READMEs etc is one of my greatest passion, I like it a lot!)

## Changes

- Added a list of packages to the main README.md file
- Added a README.md file to every package that didn't have one and added a folder structure schema to `language-server`
- Rewrote the VS Code extension README that appears in the VS Code marketplace
- Fix consistency issues among the `package.json` of the different packages
- Misc fixes to capitalization of file names and other stuff

Some of the changes were inspired by [the matching Svelte repo](https://github.com/sveltejs/language-tools) and other popular extensions in VS Code

#### Writing style

The writing style is fairly different from the other repos, I tried to match as much as I can but I'm unfortunately not that great at that. I'm definitely welcome to comments on that

#### Putting the steps needed to setup your development setup in another file (`CONTRIBUTING.md`)

I did this in order to match the main Astro repo. While part of me do think that it's easier when it's in the main file, consistency is more important, especially as the number of different repo Astro own gets higher

#### The structure of the repo might change "soon"

[This issue](https://github.com/withastro/language-tools/issues/116) potentially implies a lot of changes to the structure of the repo. Once that issue is in progress, I'll make changes accordingly (I'd love to contribute to that issue so it might happens at the same time)

#### Did not change the steps for development

I decided that it wasn't in the scope of this PR, in the future I'd like to make improvements there too but I think it's good enough for now. (Especially for users who are used to working on VS Code extensions, I think it'll feel natural. The screenshots are very helpful too)

## Testing

Fully tested those .md files with my eyes 😁

## Docs

It's all documentation


package.json
```diff
@@ -8,7 +8,6 @@
   },
   "scripts": {
     "release": "yarn build && changeset publish",
-    "benchmark": "yarn workspace astro run benchmark",
```
Princesseuh, line 11: Not really in the scope of this PR but this script isn't relevant to this project

packages/vscode/package.json
```diff
@@ -9,7 +9,7 @@
     "theme": "dark"
   },
   "version": "0.8.8",
-  "author": "Astro",
```
Princesseuh, line 12: I changed it here too but I'm not sure if it can cause issues with publishing? I checked the VS Code documentation regarding this and it said it wouldn't cause any issues though

natemoo-re: Shouldn't cause any problems! Code uses the `publisher` key, which is "astro-build".

natemoo-re (approved):
This is awesome and very helpful, thanks!

jonathantneal (approved):

jonathantneal:
This is really helpful. Thank you, @Princesseuh!

PR #160, opened 2022-01-26
Add EditorConfig, ESLint and Prettier config files from main repo

Princesseuh:
## Changes

This adds the config files used by ESLint, Prettier, EditorConfig from the main repo to this repo as well, in order to promote consistent formatting and linting across Astro project

I removed a few things from the ignores files that were applicable only to the main repo. Additionally, despite the plugins used for Prettier and ESLint, I don't think any new dependencies are needed as I think they were already installed

## Testing

No tests needed

## Docs

No docs needed

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,15 @@
+# https://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_size = 2
+indent_style = tab
+insert_final_newline = true
+trim_trailing_whitespace = false
+
+[{.*,*.md,*.json,*.toml,*.yml,}]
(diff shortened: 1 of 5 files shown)


matthewp (approved):
Looks great!

jonathantneal (approved):
This is great! Thanks, @Princesseuh. The next step would be to apply these to the repo itself?

Princesseuh:
> This is great! Thanks, @Princesseuh. The next step would be to apply these to the repo itself?

I'm thinking that we could do a commit applying formatting change (adding a `.git-blame-ignore-revs` file in the process probably) for now and after that, we could add the formatting and linting to CI like it is on the main repo?

jonathantneal:
Perfect. Merging this.

PR #161, opened 2022-01-28
Migrate to Turbo and cleanup dependencies

Princesseuh:
## Changes

This migrate the repo from Lerna to Turbo, similarly to how it was done in the main repo. Since they work similarly and all of the scripts are wrapper over it, we don't need any updates to the CI scripts (if everything goes well, computers are complicated)

Additionally, I cleaned up the dependencies, removing those that are unused. I realize that in the future, some of those might get brought back (for tests or other reasons) but I figured that for now, it was just confusing having them there and made the installation longer than it should (this change was motivated by me making this PR on coffee shop wifi)

## Testing

Tests still pass (as few as they are), additionally I tested manually by running the different scripts and everything worked

## Docs

Just replaced mention of Lerna with Turbo in the different .md

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ dist/
 _site/
 *.log
 package-lock.json
+.turbo/
 
 # .vscode files other than at top-level
 **/.vscode
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,7 +4,7 @@
 
 ## Setup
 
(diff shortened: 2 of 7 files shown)


matthewp:
YES! Looks great :D

matthewp (approved):
lgtm

jonathantneal (approved):
So much red. This is great!

Issue #155, opened 2022-01-18
🐛 BUG: If a variable starts with source and is compared with <, an error is shown


GrafZ3pp3lin:
### Describe the Bug

This error occurs in vscode with the Astro extension.

If a variable in the frontmatter script starts with source and is compared with the smaller symbol (<) and later at some point a larger symbol (>) comes, an error is displayed in vscode.

![image](https://user-images.githubusercontent.com/40472792/149999646-b0edbeb9-ada4-4699-8079-c2412fef4a8b.png)
![image](https://user-images.githubusercontent.com/40472792/150000750-43b05123-9ab8-4dfb-9441-5eb0d5ced7f3.png)

### Steps to Reproduce

1. `npm init astro -- --template minimal`
2. start a variable with source in the frontmatter script
3. Compare the variable with <
4. Use somewhere in the Frontmatter script the > symbol
5. Error!

Princesseuh:
This is caused by [this line](https://github.com/withastro/language-tools/blob/2910b038e1408940c6ecc8b9217d9d5684598ef2/packages/language-server/src/plugins/typescript/astro2tsx.ts#L50) as it fails with any of the cases in it (`hr`, `meta` etc)

Will see what I can do, thanks for the report!

PR #162, opened 2022-01-29
Apply relevant content transformations for TSX generation only in relevant parts

Princesseuh:
## Changes

Before, transformations (also knowns as, `.replace`s) used when generating TSX from `.astro` files were applied to the entire file, which meant that sometimes, we'd get false positive where the regexes thinks they're transforming HTML but really they were operating on the frontmatter content. I think that in the past, this might have been allowable because we could have JSX in the frontmatter, but this isn't the case anymore 

This fixes https://github.com/withastro/language-tools/issues/155 though I reckon that this probably caused other undocumented issues for other people in the past

I realize that this is a bit of a crude fix, but then again, the original code is already a bit brittle (not that there's necessarily another way to do it without making a whole thing). In the future, I'd assume we'd use the compiler for this. This is more so a temporary fix

## Testing

Tested manually using the snippet shared in the previously mentioned issue

## Docs

No docs needed

### DIFF
--- a/packages/language-server/src/plugins/typescript/astro2tsx.ts
+++ b/packages/language-server/src/plugins/typescript/astro2tsx.ts
@@ -23,70 +23,74 @@ interface Astro2TSXResult {
   code: string;
 }
 
-export default function(content: string): Astro2TSXResult {
-  let result: Astro2TSXResult = {
-    code: ''
-  };
+export default function (content: string): Astro2TSXResult {
+	let result: Astro2TSXResult = {
+		code: '',
+	};
 
-	const astroDocument = parseAstro(content)
+	const astroDocument = parseAstro(content);
 
 	// Frontmatter replacements
-  let frontMatterRaw = content
-		// Handle case where semicolons is not used in the frontmatter section
-		.replace(/((?!^)(?<!;)\n)(---)/g, (_whole, start, _dashes) => {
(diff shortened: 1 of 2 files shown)


packages/language-server/src/plugins/typescript/astro2tsx.ts
```diff
@@ -27,13 +28,19 @@ export default function(content: string): Astro2TSXResult {
...
   };
 
-  // Replace frontmatter marks with comments
-  let raw = content
-    // Handle case where semicolons is not used in the frontmatter section
-    .replace(/((?!^)(?<!;)\n)(---)/g, (_whole, start, _dashes) => {
-      return start + ';' + '//';
-    })
-    .replace(/---/g, '///')
+	const astroDocument = parseAstro(content)
```
Princesseuh, line 31: Not sure if using `parseAstro` is allowed here. I first tried to do this by passing `this.doc.astro` to `astro2tsx` in `DocumentSnapshot.ts` but sometimes it returned undefined and it made the extension crash. 

I figure that this cause a double parse of the file but I wasn't sure how to do otherwise

Princesseuh:
Uh oh, tests are failing. I'll investigate more

matthewp:
Looks good, thanks!

Issue #173, opened 2022-03-01
🐛 BUG: Regression: Errors are appearing on the wrong line


Princesseuh:
### Describe the Bug

![image](https://user-images.githubusercontent.com/3019731/156186943-a0d32edb-b875-42cf-a92c-f6d4db32c815.png)

This error is showing on the wrong line since the latest update (0.8.9), this is probably due to the changes to JSX generation in https://github.com/withastro/language-tools/pull/162

### Steps to Reproduce

Create any kind of .astro file with the following content:

```astro
---
---

<Component />
```

Error will be on the wrong line

PR #174, opened 2022-03-01
Fix missing whitespace in generated TSX

Princesseuh:
## Changes

Previously we were making up our TSX's template content from where the template part started in the .astro files, however what this meant is that we were missing crucial whitespace between the frontmatter and the template part. So now, we start our HTML from where the frontmatter end and not where the content starts

This fixes https://github.com/withastro/language-tools/issues/173

## Testing

Tested manually and ran tests locally

image

(Error now appears on the right line no matter the amount of white lines before)

## Docs

Bug fix only

### DIFF
--- a/.changeset/heavy-dodos-brake.md
+++ b/.changeset/heavy-dodos-brake.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/language-server': patch
+---
+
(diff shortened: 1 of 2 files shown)


matthewp (approved):
lgtm

Issue #171, opened 2022-02-28
🐛 BUG: Changing the tsconfig doesn't reflect in the editor


matthewp:
### Describe the Bug

If you make a change to the tsconfig.json file, the changes are not reflected by the language server.

It would be nice if we could watch for changes to this file and redo diagnostics.

### Steps to Reproduce

Change the tsconfig.json file, see if the changes are reflected.

PR #175, opened 2022-03-01
Add support for reloading the language server on Astro / TS config changes

Princesseuh:
## Changes

This add support for reloading the language server whenever `tsconfig.json` (and other similar files) is modified. I also added a command to manually reload the language server. This is helpful for whenever users encounter an issue where just reloading fix it (this help avoid needing to reload the entire VS Code window, which is not always pleasant to do)

The commit is a bit bigger than it should be because of formatting changes and also a small refactor removing unnecessary wrappers and abstractions. See GitHub comments on code for more info on why some changes were made

This fixes https://github.com/withastro/language-tools/issues/171

## Testing

Tested manually as our tests don't cover this (not sure if it's even possible to test a VS Code extension)

## Docs

No docs needed, the added command is self explanatory

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -13,7 +13,7 @@ module.exports = {
     "@typescript-eslint/no-use-before-define": "off",
     "@typescript-eslint/no-var-requires": "off",
     "@typescript-eslint/no-this-alias": "off",
-    "no-console": "warn",
+    "no-console": ["error", { allow: ["warn", "error"] }],
     "no-shadow": "error",
     "prefer-const": "off",
     // 'require-jsdoc': 'error', // re-enable this to enforce JSDoc for all functions
--- a/packages/vscode/src/index.ts
+++ b/packages/vscode/src/index.ts
@@ -1,71 +1,103 @@
-import * as vscode from 'vscode';
-import * as lsp from 'vscode-languageclient/node.js';
+import { window, commands, workspace, ExtensionContext, TextDocument, Position } from 'vscode';
+import { LanguageClient, RequestType, TextDocumentPositionParams, ServerOptions, TransportKind } from 'vscode-languageclient/node';
+import { LanguageClientOptions } from 'vscode-languageclient';
 import { activateTagClosing } from './html/autoClose.js';
 
-let docClient: lsp.LanguageClient;
+const TagCloseRequest: RequestType<TextDocumentPositionParams, string, any> = new RequestType('html/tag');
 
-const TagCloseRequest: lsp.RequestType<lsp.TextDocumentPositionParams, string, any> = new lsp.RequestType('html/tag');
+export async function activate(context: ExtensionContext) {
+	const serverModule = require.resolve('@astrojs/language-server/bin/server.js');
 
-/**  */
-export async function activate(context: vscode.ExtensionContext) {
-  docClient = createLanguageService(context, 'doc', 'astro', 'Astro', 6040);
+	const port = 6040;
+	const debugOptions = { execArgv: ['--nolazy', '--inspect=' + port] };
 
-  await docClient.onReady();
+	const serverOptions: ServerOptions = {
+		run: { module: serverModule, transport: TransportKind.ipc },
+		debug: {
+			module: serverModule,
+			transport: TransportKind.ipc,
+			options: debugOptions,
+		},
+	};
+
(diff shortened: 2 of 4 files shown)


okikio:
Does this change (check for language server config) allow the Astro extension to work in a web environment?

Princesseuh:
> Does this change (check for language server config) allow the Astro extension to work in a web environment?

I'm not sure I understand what you mean? Are you asking if this change is compatible with Astro running in web environnements?

okikio:
> > Does this change (check for language server config) allow the Astro extension to work in a web environment?
> 
> I'm not sure I understand what you mean? Are you asking if this change is compatible with Astro running in web environnements?

Yes

.eslintrc.cjs
```diff
@@ -13,7 +13,7 @@ module.exports = {
     "@typescript-eslint/no-use-before-define": "off",
     "@typescript-eslint/no-var-requires": "off",
     "@typescript-eslint/no-this-alias": "off",
-    "no-console": "warn",
+    "no-console": ["error", { allow: ["warn", "error"] }],
```
Princesseuh, line 16: I changed this because using `console.error` or `warn` isn't necessarily invalid in the context of this project. Right now both usages of `console.log` could be removed so this shouldn't be a problem

packages/vscode/src/index.ts
```diff
@@ -1,71 +1,103 @@
-import * as vscode from 'vscode';
-import * as lsp from 'vscode-languageclient/node.js';
+import { window, commands, workspace, ExtensionContext, TextDocument, Position } from 'vscode';
+import { LanguageClient, RequestType, TextDocumentPositionParams, ServerOptions, TransportKind } from 'vscode-languageclient/node';
+import { LanguageClientOptions } from 'vscode-languageclient';
```
Princesseuh, line 3: Changed the imports to importing specific things like this to match other files in the project

packages/vscode/src/index.ts
```diff
@@ -1,71 +1,103 @@
...
-    language: vscode.env.language,
-  };
-  const clientOptions: lsp.LanguageClientOptions = {
-    documentSelector: [{ scheme: 'file', language: 'astro' }],
-    synchronize: {
-      configurationSection: ['astro', 'javascript', 'typescript', 'prettier'],
-      fileEvents: workspace.createFileSystemWatcher('{**/*.js,**/*.ts}', false, false, false),
-    },
-    initializationOptions: {
-      ...serverInitOptions,
```
Princesseuh, line 41: Removed `serverInitOptions` from the final result since.. Well, I'm not sure what it does? Those variables are not used anywhere else and additionally, they're not in the VS Code's types for `initializationOptions`, not sure where they came from

okikio:
When we say config are we talking about `astro.config.js` or the `.vscode/settings.json`>

Princesseuh:
> When we say config are we talking about `astro.config.js` or the `.vscode/settings.json`>

The former and `tsconfig.json` (and `jsconfig.json`). The latter does not need reload since we can access the configuration directly (I believe at least, haven't checked yet)

okikio:
Then I'm not sure the extension will work in something like [`vscode.dev`](https://vscode.dev), since, vscode.dev doesn't have access to a filesystem.

Princesseuh:
> Then I'm not sure the extension will work in something like [`vscode.dev`](https://vscode.dev), since, vscode.dev doesn't have access to a filesystem.

It does through VSCode's API:

> The opened workspace or folder is on a virtual [file system](https://code.visualstudio.com/api/references/vscode-api#FileSystem). Access to workspace files needs to go through the VS Code file system API accessible at vscode.workspace.fs.

The watcher here is used through a VS Code API

okikio:
Oh, cool, nvm then

packages/vscode/src/index.ts
```diff
@@ -1,71 +1,103 @@
...
+
+		restartingClient = true;
+		await client.stop();
+
+		client = createLanguageServer(serverOptions, clientOptions);
+		context.subscriptions.push(client.start());
+		await client.onReady();
+
+		if (showNotification) {
+			window.showInformationMessage('Astro language server restarted.');
```
matthewp, line 86: Nice, I like these touches.

matthewp (approved):
Looks fantastic!

PR #178, opened 2022-03-03
Fix language server reload on config changes not working on Windows

Princesseuh:
## Changes

Previously, we splitted the path to a potential ts/jsconfig using a forward slash, I thought that was fine because I thought that VS Code normalized paths but it turns out, that: it doesn't. So we need to also split on backslashes for Windows 

## Testing

Tested manually on both Windows and Linux

## Docs

No docs needed

### DIFF
--- a/packages/vscode/src/index.ts
+++ b/packages/vscode/src/index.ts
@@ -57,7 +57,7 @@ export async function activate(context: ExtensionContext) {
 
 	// Restart the language server if any critical files that are outside our jurisdiction got changed (tsconfig, jsconfig etc)
 	workspace.onDidSaveTextDocument(async (doc: TextDocument) => {
-		const fileName = doc.fileName.split('/').pop() ?? doc.fileName;
+		const fileName = doc.fileName.split(/\/|\\/).pop() ?? doc.fileName;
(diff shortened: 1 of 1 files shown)


packages/vscode/src/index.ts
```diff
@@ -57,7 +57,7 @@ export async function activate(context: ExtensionContext) {
 
 	// Restart the language server if any critical files that are outside our jurisdiction got changed (tsconfig, jsconfig etc)
 	workspace.onDidSaveTextDocument(async (doc: TextDocument) => {
-		const fileName = doc.fileName.split('/').pop() ?? doc.fileName;
+		const fileName = doc.fileName.split(/\/|\\/).pop() ?? doc.fileName;
```
matthewp, line 60: This is probably a good case to use path.sep: https://nodejs.org/api/path.html#pathsep

Princesseuh: Unfortunately, I believe `path` is not available for web extensions in this context:
> Node.js globals and libraries such as process, os, setImmediate, path, util, url are not available at runtime

matthewp: Ah, ok that make sense.

matthewp (approved):

PR #181, opened 2022-03-03
Disable publishing to open-vsx

matthewp:
## Changes

- Disables open-vsx publishing for now, as we do not have the `astro-build` publisher, that we need.

## Testing

N/A

## Docs

N/A


Princesseuh (approved):
lgtm

Issue #81, opened 2021-09-23
🐛 BUG: Emmet doesn't work inside a component


RafidMuhymin:
### Describe the Bug

Emmet doesn't work inside a component

### Steps to Reproduce

Emmet should work inside components

SeizeDub:
Same issue here

Princesseuh:
I'm not sure what you mean by inside a component? For me Emmet works in both components files and inside slots

RafidMuhymin:
@Princesseuh The below GIF may help you understand the issue better

![Peek 2022-03-04 15-44](https://user-images.githubusercontent.com/63650415/156739877-a990ff71-1583-491e-84ae-15ce8a9fa45f.gif)

Princesseuh:
@RafidMuhymin  Hmm, for me it works in slots, I'll investigate what's happening, thank you!

PR #183, opened 2022-03-04
Add support for Emmet inside components, upgrade version

Princesseuh:
## Changes

This add support for Emmet inside components slots, in most cases users are using slots for HTML so this is really helpful. 

Additionally, this PR upgrade Emmet to the latest version (well, it upgrade the vscode helper which in turns upgrade Emmet). This fixes a bunch of issues in the background and add a few missing completions that were adding over time. 

This also remove the annoying message you got every time you install the language-server (which affect the main astro project due to the check command) where it tells you that vscode-emmet-helper is deprecated

Fixes https://github.com/withastro/language-tools/issues/81

## Testing

Tested manually

## Docs

No docs needed

### DIFF
--- a/packages/language-server/package.json
+++ b/packages/language-server/package.json
@@ -25,7 +25,7 @@
     "ts-morph": "^12.0.0",
     "typescript": "^4.5.4",
     "vscode-css-languageservice": "^5.1.1",
-    "vscode-emmet-helper": "2.1.2",
+    "@vscode/emmet-helper": "^2.8.4",
     "vscode-html-languageservice": "^3.0.3",
     "vscode-languageserver": "6.1.1",
     "vscode-languageserver-protocol": "^3.16.0",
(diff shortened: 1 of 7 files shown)


packages/vscode/package.json
```diff
@@ -26,8 +26,6 @@
   "dependencies": {
     "@astrojs/language-server": "0.9.0",
     "@astrojs/ts-plugin": "0.2.0",
-    "vscode-emmet-helper": "2.1.2",
-    "vscode-html-languageservice": "^3.0.3",
```
Princesseuh, line 30: Those are unneeded, we don't use those in the extension but in the language-server, this created duplicate dependencies

packages/language-server/package.json
```diff
@@ -25,7 +25,7 @@
     "ts-morph": "^12.0.0",
     "typescript": "^4.5.4",
     "vscode-css-languageservice": "^5.1.1",
-    "vscode-emmet-helper": "2.1.2",
+    "@vscode/emmet-helper": "^2.8.4",
```
matthewp, line 28: Nice, do these add new features?

matthewp (approved):
lgtm

packages/language-server/package.json
Princesseuh: They don't have a changelog but as far as I could tell from the commit history, it was only fixes

PR #186, opened 2022-03-06
feat: added icon to language

JuanM04:
## Changes

- What does this change?


JuanM04:
I've taken the icon from here: https://github.com/vscode-icons/vscode-icons/pull/2774

antonyfaris:
> I've taken the icon from here: [vscode-icons/vscode-icons#2774](https://github.com/vscode-icons/vscode-icons/pull/2774)

The viewbox property has a different value. Maybe using the original size (32x32) will fix it?

Princesseuh:
Unfortunately doesn't seem to change anything: 
image

JuanM04:
Ok, I've found [this icon](https://github.com/withastro/astro/blob/main/assets/brand/file-icon.svg), maybe this one works

Princesseuh:
Unfortunately, same thing with that one haha. Does it work fine on your side? Maybe it's an issue with my particular setup

JuanM04:
I can replicate it but I don't know how to fix it. In only happens with the Seti theme, so maybe we could submit an icon there (?)

JuanM04:
It seems that we aren't the only ones with alignment issues: https://github.com/microsoft/vscode/issues/141335#issuecomment-1021761492

natemoo-re (approved):
IMO a misaligned icon is better than no icon! I'd merge this and we can fix the alignment later?

JuanM04:
@natemoo-re Let me do a quick try by removing the margins with figma. If it stays the same, I'll merge it

JuanM04:
Plot twist: it didn't work

Issue #142, opened 2021-12-30
🐛 BUG: Import svelte or vue files


vhladko:
### Describe the Bug

If you are using latest templates and in your tsconfig file has:
```
compilerOptions: {
  moduleResolution: "node",
}
```
And try to import `svelte` or `vue` components to your astro component, you will get and error from extension.

### Steps to Reproduce

1. `npm init astro` using template vue/svelte
2. import svelte or vue component into your astro component
3. Error! Astro extension will show you an error 
![Screenshot 2021-12-30 at 10 11 14](https://user-images.githubusercontent.com/56257437/147733439-15960b78-6dc7-46bd-ad07-ca1c77559acb.png)

dslatkin:
Also had the same issue! The `.tsconfig` mentioned here that comes with the default Svelte template seemed to be causing it, removing the file (presumably meaning I'm going with just JS for now) seems to fix the issue. Thanks to Erika and Muhymin in the Discord for the help!

jonathantneal:
This is as far as I got:
- Svelte files in the frontmatter of an Astro file are not being run thru the checks, like `isVirtualSvelteFilePath`. Maybe that’s how it’s supposed to work.
- I was able to narrow down that the checks might be happening in this file  (`packages/language-server/src/plugins/typescript/features/DiagnosticsProvider.ts`) maybe in `getDiagnostics`.

Issue #126, opened 2021-12-03
🐛 BUG: Type errors aren't shown for imported interfaces and types


RafidMuhymin:
### Describe the Bug

The VSCode extension doesn't display type errors for imported interfaces and types.
![Peek 2021-12-03 23-24](https://user-images.githubusercontent.com/63650415/144651211-59fe832a-ed99-4600-bee1-8a51ca36a4fe.gif)

### Steps to Reproduce

The VSCode extension should display type errors for imported interfaces and types.

Princesseuh:
This is a TypeScript issue, TypeScript expects you to not put any file extension (or .js) when importing TS files

Nonetheless, since Astro recommend (or at least used to) to put the file extension, we should make sure that it works even when you put the extension, good catch

Princesseuh:
Now that the extension recommend TypeScript compatible imports, this shouldn't be a problem anymore. Closing

Issue #106, opened 2021-11-07
🐛 BUG: The types of the imported components aren't updated 


RafidMuhymin:
### Describe the Bug

The types of the imported components aren't updated. I'm getting the same TypeErrors when I update the types of a component's prop.

### Steps to Reproduce

The types of the imported components should be updated as soon as the user updates them

PR #197, opened 2022-03-09
Add support for importing and parsing files from frameworks

Princesseuh:
## Changes

I first want to start this by giving huge thanks to @matthewp for helping me out with this because between the Document, DocumentSnapshot, DocumentFragmentSnapshot and other Java-like classes I was (and still fairly is) lost most of the time, he really helped me out figure out how to do this!

This PR adds support for importing Svelte and Vue files and for Svelte files especially, converting them to TSX and getting proper props information, like we do for Astro files. 

This setup a bigger plan to eventually allow users to add support for any kind of framework that doesn't use JSX/TSX files without needing to make changes to the language-server itself. We're not there just yet (and we probably want to make things a lot cleaner before we do that) but, this is a first step

Additionally, this PR fix needing to remove the extension from `.ts` file in order to get types and completions. Technically, it is not wrong from TypeScript to require the extension to be removed as it is its normal behavior. However, Astro (by the past at least) as always recommended to put the extension, so we'll match that here. The way it works on the technical side is that we pretend that `.ts` files are `.tsx` files. I'm not sure why it works, but it does

I am not proud of the code but, it works and it is an improvement. Notably, the error when importing Svelte and Vue files was one of the most mentioned issues in our support threads last month

Fixes https://github.com/withastro/language-tools/issues/142 https://github.com/withastro/language-tools/issues/126

#### Diagnostics and hover information is currently unavailable for imported framework components

Not sure why, my guess is that TypeScript doesn't get the DocumentSnapshot of our Svelte file. I'm not sure why, it should technically be going through our SnapshotManager and get the proper Snapshot for said file but it seems like it doesn't. Completions however are working nicely because they go through a different system (which maybe diagnostics could also use, not sure)

#### Types from imported Vue and Svelte files are not updated when those files are changed

Again not sure why, I think this is part of a potentially bigger issue related to cache (ex https://github.com/withastro/language-tools/issues/106) or maybe it's just a question of getting the language-server to update its cache when those files are updated

## Testing

Tested manually for now. When I figure out how to make diagnostics works (in a later PR), I plan to also write tests for diagnostics for different framework's components

## Docs

Technically a bug fix so no docs needed

### DIFF
--- a/packages/language-server/package.json
+++ b/packages/language-server/package.json
@@ -19,7 +19,7 @@
     "test": "tap --no-coverage test/**"
   },
   "dependencies": {
-    "@astrojs/svelte-language-integration": "*",
+    "@astrojs/svelte-language-integration": "^0.1.0",
     "@vscode/emmet-helper": "^2.8.4",
     "lodash": "^4.17.21",
     "source-map": "^0.7.3",
--- a/packages/language-server/package.json
+++ b/packages/language-server/package.json
@@ -10,7 +10,6 @@
   },
   "files": [
     "dist",
-    "astro.d.ts",
     "bin",
     "types"
   ],
@@ -20,6 +19,7 @@
     "test": "tap --no-coverage test/**"
   },
   "dependencies": {
+    "@astrojs/svelte-language-integration": "*",
     "@vscode/emmet-helper": "^2.8.4",
     "lodash": "^4.17.21",
     "source-map": "^0.7.3",
--- a/.changeset/smooth-candles-rhyme.md
+++ b/.changeset/smooth-candles-rhyme.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/language-server': minor
+---
+
+Add support for loading files from non-JSX frameworks such as Vue and Svelte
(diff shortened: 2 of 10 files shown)


packages/language-server/package.json
```diff
@@ -20,6 +19,7 @@
     "test": "tap --no-coverage test/**"
   },
   "dependencies": {
+    "@astrojs/svelte-language-integration": "*",
```
matthewp, line 22: Shouldn't this be a semver range?

matthewp, line 22: Shouldn't this be a semver range?

Princesseuh: I wasn't really sure on how to proceed for an internal package like this, I hesitated with `latest` but I read somewhere that `latest` sometimes has issues in workspaces. We could just put the version of the package, for sure

matthewp: I think that's what we want, look at how we do it in the main `astro` package for example. Changesets will handle updating the version when it changes.

Issue #202, opened 2022-03-11
🐛 BUG: CSS completions only work in the first style tag in a document


Princesseuh:
### Describe the Bug

This is caused by a Svelte-ism in the way we recuperate the content of style tags. Since in Svelte you can only have one style tag per component, this is not an issue but in Astro files, you can realistically have multiple style tags 

The [offending code is (partly) here](https://github.com/withastro/language-tools/blob/fd92a856293febb71079a2163842ca7f7d462f0d/packages/language-server/src/core/documents/utils.ts#L255-L263). 

Probably what should happen is that our document should contain a list of style tags (array of TagInformation) and whenever we do operation in our CSSPlugin, we first check in which style tag we are and then give appropriate responses (we can have multiple style tags with different langs in a single page and thus needing different completions depending on that for instance)

### Steps to Reproduce

1. Create an Astro file with multiple style tag
2. Get completions in the first one
3. But not in the second one (or any after)!

matthewp:
Oh good catch!

Issue #201, opened 2022-03-10
📘 DOC: Write internal docs


Princesseuh:
Despite being what might seems to be a simple project, this repo is fairly hard to navigate and understand. It's not really always clear what the different packages and parts of the code are doing and how they interact with each others. I've had multiple users asks me how it works because they were interested in contributing and just.. didn't know where to start (which is clearly understandable, even after working with it for a few months now, I still don't really know what's happening all the time :sweat_smile: )

A first step in making this clearer was my first PR to this repo (https://github.com/withastro/language-tools/pull/159), which, I like to think improved the situation a bit but we can do even better! 

In this issue, I suggest that we write that we write internal documentation, similar to [the one Svelte has for its language-tools](https://github.com/sveltejs/language-tools/tree/master/docs/internal) in addition of improving comments in the project code itself in order to make it much easier for new contributors to jump in and understand the repo

Issue #180, opened 2022-03-03
💡 RFC: Proper web environnement support


Princesseuh:
### Background & Motivation

Right now when installing the Astro extension in a web VS Code environnement such as [vscode.dev](https://vscode.dev) you.. don't get much outside of syntax coloring because the extension isn't packaged properly and is not following the requirements for web extensions

### Possible solutions

Follow the requirements at [https://code.visualstudio.com/api/extension-guides/web-extensions](https://code.visualstudio.com/api/extension-guides/web-extensions) and see how well we fare

### Help make it happen!

- [X] I am willing to submit a PR to implement this change.
- [ ] I am willing to submit a PR to implement this change, but would need some guidance.
- [ ] I am not willing to submit a PR to implement this change.

Princesseuh:
Interesting repo to check when we'll attack this: https://github.com/asafamr/svelte-vscode-web

PR #209, opened 2022-03-15
Refactor language-server package

Princesseuh:
## Changes

This is a refactor of the language-server package. Do not be fooled by the large amount of changes as this PR also finally apply the formatting changes added in https://github.com/withastro/language-tools/pull/160 which makes for a large portion of the changes.

A lot of the changes especially in the TypeScript plugin are inspired by updates that were made to the Svelte language-server since we forked off them, notably, a lot of the changes they've done in the last year or so are performance related which by proxy also happens to fix a few cache issues we've been having. Huge credits to the Svelte team for their work, always very inspiring. Some changes were also inspired by [Volar](https://github.com/johnsoncodehk/volar), another very impressive language-server codebase to look at

Certain parts were not touched much (if at all) because they'll be subject to planned changes in the near future, such as the AstroPlugin and the Astro Check command. Work on those will be done in dedicated PRs

No new features were added apart from a completion for the `:global` selector in CSS because I noticed that Svelte had it and we didn't. It felt pretty inoffensive to add since we already supported adding completions for selectors

Due to architectural changes, this fixes https://github.com/withastro/language-tools/issues/202
An initial draft of an internal documentation was added, which close https://github.com/withastro/language-tools/issues/201, further work on it will be in dedicated issues to specific subjects
The new VS Code test added fix https://github.com/withastro/language-tools/issues/210
I believe that was already fixed before but even then, this fixes https://github.com/withastro/language-tools/issues/106 (and also does it for non-astro components and js/ts files)

Thanks to the changes in the entrypoints, this setup the way for https://github.com/withastro/language-tools/issues/180 (but we are not quite there yet)

## Testing

New tests setups were added for both the extension itself and the language-server, instead of using the `astro check` command, we now directly test the language-server itself using a similar setup to the one we use in the main repo (Mocha + Chai).

Unlike the previous tests and the main repo, I wrote the tests using TypeScript because it is in my opinion really hard to work with the VS Code packages without types due to the many many different types with similar shapes and names

Additional testing was done manually using various Astro projects

## Docs

Internal documentation was added in the `docs` folder, no docs should be needed on user side


.github/workflows/ci.yml
```diff
@@ -52,5 +52,10 @@ jobs:
       - name: Build the packages
         run: yarn build
 
+      - name: Test (Linux)
+        run: xvfb-run -a yarn test
```
Princesseuh, line 56: This is necessary for the VS Code tests, see [this page](https://code.visualstudio.com/api/working-with-extensions/continuous-integration#github-actions)

packages/language-server/src/node.ts
```diff
@@ -0,0 +1,6 @@
+import * as vscode from 'vscode-languageserver/node';
+import { startLanguageServer } from './server';
+
+const connection = vscode.createConnection(vscode.ProposedFeatures.all);
+
+startLanguageServer(connection);
```
matthewp, line 6: Just curious, is having this in its own file setting up to better support non-Node environments?

matthewp (approved):
Looks great!

packages/language-server/test/check/diagnostics.test.ts
```diff
@@ -18,7 +18,7 @@ describe('Astro Check', async () => {
 		const [{ diagnostics }] = await checker.getDiagnostics();
 
 		expect(diagnostics.length, 'Expected diagnostics to be empty').to.equal(0);
-	});
+	}).timeout(10000);
```
Princesseuh, line 21: I'm not sure why but the first test in this file is taking a very long time. The tests in this file are kinda faulty in general and will be removed in the future (we don't need to test through AstroCheck anymore)

PR #214, opened 2022-03-17
Fix errors with malformed Svelte components, fix missing libs in some context

Princesseuh:
## Changes

Wrap the call to `svelte2tsx` in a try block in order to avoid the extension crashing because of a malformed Svelte component. Instead, when it doesn't work it just returns a default empty function like we do for Vue components. I alternatively tried to set the `emitOnTemplateError` to `true` but it didn't seems to work (and wouldn't be a really reliable solution anyway). 

In the future, a better fix would be to return the error from the Svelte compiler to users directly in the Astro file, that'd be neat

This also removes the forcing of `lib: ['ESNext'],` in the TypeScript config for Astro files as it didn't work in certain contexts, not sure why but this fixes the issue for now

## Testing

Both fixes tested manually on repos where it didn't work previously

## Docs

No docs needed


matthewp (approved):
Looks good!

Issue #168, opened 2022-02-20
🐛 BUG: Issue with types and intellisence in astro project with framework ( react - solid - preact )


KirdesMF:
### Describe the Bug

First, sorry I did not try with `svelte` and `vue` , I can still do it if this is relevant, just ask me :)

Intellisence does not work in `.tsx` file and elements are typed as `any` instead of `JSX.IntrinsicElements` , attributes are ignored ...
so it seems there is some missing types.

I did some investigation and get the correct types but this will break some other thing.

Lets resume this a bit:

`astro-solid`  => there is no need to install a type package but we need to add options to  `tsconfig.json`

```json
{
  "compilerOptions": {
    "moduleResolution": "node",
    "jsx": "preserve",  // this
    "jsxImportSource": "solid-js" // and this
  }
}

```

but doing this breaks jsx in `.astro` files 

![image](https://user-images.githubusercontent.com/54989552/154847328-ae8d3d12-25cc-4320-a415-98cc4b43cefc.png)

and I know auto import in `.astro` file does not atm, but even manually the vscode hint is wrong (it does no show the `.tsx`)

![image](https://user-images.githubusercontent.com/54989552/154848044-1b7a747f-6a47-46ff-9a36-749e29f25000.png)

and there is no error for missing props, no autocomplete

![image](https://user-images.githubusercontent.com/54989552/154848073-6abf6992-04e6-4595-9c05-ca98c4bdc95c.png)

**N.B** : this is maybe an other issue 

---

`astro-preact` => same as `astro-solid` but with this : 

```json
{
  "compilerOptions": {
    "moduleResolution": "node",
    "jsx": "react-jsx",
    "jsxImportSource": "preact"
  }
}
```
it also breaks how `jsx` is overwritten by `astro`

`astro-react` : we need to install `@types/react` but no need to edit the tsconfig, not sure if we also need to add `@types/react-dom`, I guess no because all of that is handle by the `astro renderer` ( really not sure, if someone could enlighten me <3 )

---

I dont know how all of this works under the hood, but maybe  in the `astro CLI`  when creating a project, astro could ask if we plan to use typescript or add options depending on `vite` templates.

```PowerShell
- react
- react-ts
- preact
- preact-ts
```

Do not hesitate to ask me more details or more tests

👋🏽 

### Steps to Reproduce

1. `pnpm init astro` using template react | solid | preact
2. create a component inside a `.tsx` file
3. check intellisence
4. be mad

Issue #141, opened 2021-12-26
🐛 BUG: `client:` attributes not being parsed correctly


ProspectPyxis:
### Describe the Bug

At least with the `framework-solid` template, the Astro extension seems to not parse `client` attributes in imported components correctly, giving an incorrect error. Everything works fine otherwise.

### Steps to Reproduce

1. `pnpm create astro` using template `framework-solid` (npm or yarn likely does the same thing, but I haven't tested this)
2. Open the resulting project in vscode, with only the Astro plugin enabled
3. Open the `index.astro` file
4. The line `<Counter client:visible>` throws an error:
  ```
  Type '{ "client:visible": true; }' is not assignable to type '{ children: any; }'.
    Property 'client:visible' does not exist on type '{ children: any; }'.js(2322)
  ```

ProspectPyxis:
Update: this issue seems to only happen with `.jsx` files. After simply renaminng `Counter.jsx` to `Counter.tsx` with no other changes to the file and changing the `import Counter` statement appropriately, the issue disappears. Thankfully, I'm planning to code in typescript, so this issue won't affect me in this case, but I'll leave this open for investigation in case other people runs into the same issue and wishes to use javascript.

Simperfy:
I'm experiencing the same issue as well.

snelgrove:
I'm using Typescript (and Svelte, if that's relevant) and I get a very similar message :
```ts
Type '{ "client:load": true; list: any; }' is not assignable to type 'IComponentOptions<Record<string, any>>'.
  Property 'client:load' does not exist on type 'IComponentOptions<Record<string, any>>'.
```
No idea why, but the error seems to disappear if I delete `tsconfig.json`.

PoQuatre:
This issue is happening to me for `.tsx` files while strict mode is enabled in `tsconfig.json`.
PS: I'm using Preact (with Typescript)

PR #216, opened 2022-03-18
Provide types for our TSX output

Princesseuh:
## Changes

This add a `.d.ts` file describing our TSX output, since our diagnostics are based on those types, this bring a bunch of improvements, such as better support for Astro-specific attributes (client directives etc) on all kinds of components

Additionally, this PR add support for Hover to our HTML plugin, the reason it is done in the same PR is because without it you'd get TSX hover information when hovering HTML tags, which in addition of not being pretty or informative, might confuse users. The hover info provided by `vscode-html-languageservice` is actually pretty nice, I think users will like that a lot: 

image

This fixes https://github.com/withastro/language-tools/issues/164, fixes https://github.com/withastro/language-tools/issues/141, fixes https://github.com/withastro/language-tools/issues/217, fixes https://github.com/withastro/language-tools/issues/168 (at least the part where other typings would poison our output, the rest of that issue is not related to the extension itself, but after tihs PR we'll be able to provide better defaults in our starters without worrying about poisoning issues)

## Testing

Added tests for HTML hover info, the rest was tested manually for now

## Docs

Basic documentation was added for hover information on Astro-specific stuff however, it could use further work. I plan to collab with the docs team on this to make sure we provide info that is similarly worded to the documentation

### DIFF
--- a/packages/language-server/src/core/documents/AstroDocument.ts
+++ b/packages/language-server/src/core/documents/AstroDocument.ts
@@ -1,4 +1,4 @@
-import { HTMLDocument } from 'vscode-html-languageservice';
+import { HTMLDocument, Range } from 'vscode-html-languageservice';
 import { urlToPath } from '../../utils';
 import { WritableDocument } from './DocumentBase';
 import { AstroMetadata, parseAstro } from './parseAstro';
@@ -29,7 +29,12 @@ export class AstroDocument extends WritableDocument {
 		this.updateDocInfo();
 	}
 
-	getText(): string {
+	getText(range?: Range | undefined): string {
+		if (range) {
+			const start = this.offsetAt(range.start);
+			const end = this.offsetAt(range.end);
+			return this.content.substring(start, end);
+		}
 		return this.content;
 	}
 
--- a/packages/language-server/src/plugins/html/features/astro-attributes.ts
+++ b/packages/language-server/src/plugins/html/features/astro-attributes.ts
@@ -1,0 +1,130 @@
+import { newHTMLDataProvider } from 'vscode-html-languageservice';
+
+export const astroAttributes = newHTMLDataProvider('astro-attributes', {
+	version: 1,
+	globalAttributes: [
+		{
+			name: 'class:list',
+			description: 'Utility to provide a list of class',
+		},
+		{
+			name: 'set:html',
+			description: 'Inject unescaped HTML into this tag',
+			references: [
+				{
+					name: 'Astro documentation',
+					url: 'https://docs.astro.build/en/migrate/#deprecated-unescaped-html',
+				},
+			],
+		},
+		{
+			name: 'set:text',
+			description: 'Inject escaped text into this tag',
+			references: [
+				{
+					name: 'Astro documentation',
+					url: 'https://docs.astro.build/en/migrate/#deprecated-unescaped-html',
+				},
+			],
+		},
+	],
+	tags: [
+		{
+			name: 'script',
+			attributes: [
+				{
+					// The VS Code tag definitions does not provide a description for the deprecated `charset` attribute on script tags
+					// Which mean that since we get no hover info for this, we instead get JSX hover info. So we'll just specify a description ourselves for this specific case
+					name: 'charset',
+					description:
+						"(Deprecated) It's unnecessary to specify the charset attribute, because documents must use UTF-8, and the script element inherits its character encoding from the document.",
+				},
+				{
+					name: 'define:vars',
+					description: 'Passes serializable server-side variables into a client-side script element',
+					references: [
+						{
+							name: 'Astro documentation',
+							url: 'https://docs.astro.build/en/guides/styling/#variables-in-scripts--styles',
+						},
+					],
+				},
+			],
+		},
+		{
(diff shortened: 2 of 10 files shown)


packages/language-server/src/plugins/typescript/features/DiagnosticsProvider.ts
```diff
@@ -233,41 +233,5 @@ function isNoMarkdownBlockQuoteWithinMarkdown(
...
-	if (diagnostic.code === 1184) {
-		return {
-			...diagnostic,
-			message:
-				diagnostic.message + '\nIf this is a declare statement, move it into <script context="module">..</script>',
-		};
-	}
-
-	return diagnostic;
+	return diagnostic
```
Princesseuh, line 236: This was one of the few Svelte-ism left in the code base after the refactor, I kept the method there however because we'll use it for Astro-ism now!

packages/language-server/src/core/documents/AstroDocument.ts
```diff
@@ -29,7 +29,12 @@ export class AstroDocument extends WritableDocument {
 		this.updateDocInfo();
 	}
 
-	getText(): string {
+	getText(range?: Range | undefined): string {
+		if (range) {
+			const start = this.offsetAt(range.start);
+			const end = this.offsetAt(range.end);
+			return this.content.substring(start, end);
+		}
```
Princesseuh, line 37: Despite us not using this, it's necessary that we support it because `vscode-html-languageservice` uses that feature to get the name of the hovered HTML attribute (despite having access to it in much simpler ways in my opinion but oh well) for hover info

packages/language-server/src/plugins/html/features/astro-attributes.ts
```diff
@@ -0,0 +1,130 @@
...
+			name: 'class:list',
+			description: 'Utility to provide a list of class',
+		},
+		{
+			name: 'set:html',
+			description: 'Inject unescaped HTML into this tag',
+			references: [
+				{
+					name: 'Astro documentation',
+					url: 'https://docs.astro.build/en/migrate/#deprecated-unescaped-html',
```
matthewp, line 16: Yes! Love that you can link to the docs this way.

matthewp:
Looks very clean! Just had one question about the way we are consuming these types and if they could be done another way.

packages/language-server/src/plugins/html/features/astro-attributes.ts
Princesseuh: Yep! It's really neat. It's unfortunate that there's no way to provide this for diagnostics though, it only works for hover info

packages/language-server/src/utils.ts
```diff
@@ -38,20 +38,16 @@ export function getLastPartOfPath(path: string): string {
...
- *
- * The language service is case insensitive, and would provide
- * hover info for Svelte components like `Option` which have
- * the same name like a html tag.
+ * Return true if a specific node could be a component.
+ * This is not a 100% sure test as it'll return false for any component that does not match the standard format for a component
  */
 export function isPossibleComponent(node: Node): boolean {
-	return !!node.tag?.[0].match(/[A-Z]/);
+	return !!node.tag?.[0].match(/[A-Z]/) || !!node.tag?.match(/.+[.][A-Z]/);
```
Princesseuh, line 45: This was added to match imported component under a namespace ex: 

```astro
---
import * as thing from "component.tsx"
---

<thing.Counter/>
```

packages/language-server/src/plugins/typescript/language-service.ts
```diff
@@ -91,6 +91,14 @@ async function createLanguageService(tsconfigPath: string, docContext: LanguageS
 
 	const astroModuleLoader = createAstroModuleLoader(getScriptSnapshot, compilerOptions);
 
+	let languageServerDirectory: string;
+	try {
+		languageServerDirectory = dirname(require.resolve('@astrojs/language-server'));
+	} catch (e) {
+		languageServerDirectory = __dirname;
+	}
+	const astroTSXFile = ts.sys.resolvePath(resolve(languageServerDirectory, '../types/astro-jsx.d.ts'));
```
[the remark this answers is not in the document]
natemoo-re: I had the same question but I'm fine with this approach for now! I think it'd be good to keep these centralized in the `astro` repo, though.

packages/language-server/src/utils.ts
natemoo-re: Nice catch!

packages/language-server/src/plugins/typescript/language-service.ts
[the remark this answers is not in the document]
matthewp: Do we see issues where users don't get `astro/env` for unknown reasons?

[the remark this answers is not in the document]
matthewp: Ok, given that this is a largish change I see the argument to do it in a more conservative fashion for now and then extract later, once we've worked through bugs that may arise. This way we can tell if a bug is caused by something wrong with the type vs. the types not loading correctly.

matthewp (approved):

Issue #218, opened 2022-03-21
💡 RFC: Add hover info to CSS


Princesseuh:
### Background & Motivation

Just like https://github.com/withastro/language-tools/issues/217 but for CSS. It's the only area we don't have hover info, we should have it everywhere, it's very nice to look at!

`vscode-css-languageservice` provide this so we should be able to just ask it (nicely) from there. We need to make sure it works in multiple style tags, like completions do

### Possible solutions

Use `vscode-css-languageservice`

### Alternatives considered

N/A

### Risks, downsides, and/or tradeoffs

N/A

### Open Questions

N/A

### Help make it happen!

- [X] I am willing to submit a PR to implement this change.
- [ ] I am willing to submit a PR to implement this change, but would need some guidance.
- [ ] I am not willing to submit a PR to implement this change.

PR #219, opened 2022-03-21
Add support for getting CSS hover info

Princesseuh:
## Changes

This adds support for hover info in style tags and style attributes. Similarly to what was added for HTML in https://github.com/withastro/language-tools/pull/216

It looks like this:
image
image

This is done through `vscode-css-languageservice`, which ensure consistency with how it is done in normal `.html` and `.css` files in VS Code

This close https://github.com/withastro/language-tools/issues/218

## Testing

Added tests for this feature in addition of manual testing. Additional minor tests for other CSS features were also added in this PR to ensure consistency in how features are tested

## Docs

No docs needed

### DIFF
--- a/.changeset/chatty-garlics-jog.md
+++ b/.changeset/chatty-garlics-jog.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/language-server': patch
+---
+
+Add CSS hover info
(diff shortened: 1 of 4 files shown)


matthewp (approved):
Looks great

Issue #221, opened 2022-03-23
💡 RFC: Add support for DocumentSymbols


Princesseuh:
### Background & Motivation

Currently in an .astro files, you don't get any symbol information, which mean that the breadcrumb navigation and the outline tab are empty / do not work. It'd be cool to have this

### Possible solutions

Both the HTML and CSS language services have a method for this. For TypeScript, we can generate a list of symbols using `getNavigationTree` like Volar and Svelte does

### Alternatives considered

N/A

### Risks, downsides, and/or tradeoffs

N/A

### Open Questions

N/A

### Help make it happen!

- [X] I am willing to submit a PR to implement this change.
- [ ] I am willing to submit a PR to implement this change, but would need some guidance.
- [ ] I am not willing to submit a PR to implement this change.

PR #222, opened 2022-03-23
Add support for Document Symbols

Princesseuh:
## Changes

This add support for DocumentSymbols. I initially implemented this because I thought it would allow us to tell when a variable is deprecated and VS Code would adjust its highlighting but it turns out that it doesn't seem to use symbols for that. 

Oh well, it gives us a nice outline of the file with ability to navigate through the files using the symbols, which is really fun and nice!

image
image

Especially with the two namespaces for the Frontmatter and Template part, it makes for nice folders:

image

Fixes https://github.com/withastro/language-tools/issues/221

## Testing

Tests were added for the HTML and CSS part, rest was tested manually using various projects

## Docs

No docs needed


packages/language-server/src/plugins/typescript/features/DocumentSymbolsProvider.ts
```diff
@@ -0,0 +1,100 @@
...
+		this.collectSymbols(navTree, fragment, undefined, (symbol) => symbols.push(symbol));
+
+		const result: SymbolInformation[] = [];
+
+		// Add a "frontmatter" namespace for the frontmatter if we have a closed one
+		if (document.astroMeta.frontmatter.state === 'closed') {
+			result.push(
+				SymbolInformation.create(
+					'Frontmatter',
+					SymbolKind.Namespace,
```
Princesseuh, line 28: Used the Namespace SymbolKind here for two reasons:
- It made the most sense to me, it also look the prettiest with the {} icon
- It seems to be the one most extension uses for this kind of symbols (ex: Volar does it for the different parts of a Vue file). Svelte uses field but I think it's only because it's the default one for HTML elements (and Svelte like to consider the different parts of its files to be "HTML elements")

PR #226, opened 2022-03-23
Misc TSX typing fixes

Princesseuh:
## Changes

Add a few missing properties, fix name casing of SVG properties

## Testing

Tested manually

## Docs

No docs needed

### DIFF
--- a/packages/language-server/types/astro-jsx.d.ts
+++ b/packages/language-server/types/astro-jsx.d.ts
@@ -447,7 +447,7 @@ declare namespace astroHTML.JSX {
 		classid?: string | undefined | null;
 		cols?: number | undefined | null;
 		colspan?: number | undefined | null;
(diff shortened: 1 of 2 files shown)


packages/language-server/types/astro-jsx.d.ts
```diff
@@ -447,7 +447,7 @@ declare namespace astroHTML.JSX {
 		classid?: string | undefined | null;
 		cols?: number | undefined | null;
 		colspan?: number | undefined | null;
-		content?: string | undefined | null;
+		content?: string | URL | undefined | null;
```
Princesseuh, line 450: Some users tend to use URL for this property (especially for like og:image), since we convert those to string in Astro, it should be accepted by the language-server too

matthewp (approved):

PR #232, opened 2022-03-28
Add support for folding CSS in style tags

Princesseuh:
## Changes

Add support for folding CSS, easy enough! Much like other features powered by the VSCode's CSS language service, this only supports CSS, SCSS and LESS

image

With this change, the only thing left to be able to fold is JavaScript inside the frontmatter and script tags. That one will be a bit more complicated :sweat_smile: 

## Testing

Added tests for this, also added some missing tests for documentSymbols and foldingRanges in HTML while I was there

## Docs

No docs needed


natemoo-re (approved):
Looks awesome! Great work. ❤️

Issue #234, opened 2022-03-28
💡RFC: Add support for prop completions for JSX based frameworks


Princesseuh:
### Background & Motivation

We currently only support prop completions for Astro and Svelte completions, it'd be cool to have it for all JSX frameworks

### Proposed Solution

I am not sure why this doesn't work at the moment. What I'm thinking is that we're getting the props from the Props type instead of from the actual type being passed to the component

### Risks, downsides, and/or tradeoffs

N/A

### Open Questions

N/A

### Help make it happen!

- [X] I am willing to submit a PR to implement this change.
- [ ] I am willing to submit a PR to implement this change, but would need some guidance.
- [ ] I am not willing to submit a PR to implement this change.

PR #235, opened 2022-03-28
Improve completions for components

Princesseuh:
## Changes

This initially started as only a fix for https://github.com/withastro/language-tools/issues/234 but I kept seeing issues with completions on component and wanted them all fixed in one go :sweat_smile:

### JSX frameworks completions fixes

Previously, you'd only get completions for props on JSX-based components if they were defined using an interface. Without diving too much in TypeScript stuff, it was basically because we got the props from the defined type of the props object instead of from the object itself (TypeScript is complicated.)

This is now fixed and we provide completions for props no matter how they're defined (as long as it's a function component, at least)

### Client directives completions

Client directives completions always existed but they required you to type `client:` and only then you would get completions which isn't very good for usability or feature discovery

Since the HTML language service now (since https://github.com/withastro/language-tools/pull/216) know what client directives are, we can instead provide completions through there. We can't do this in the HTML plugin however because we need to only provide those completions on appropriate components (aka, not Astro ones) which require TS so we do this in the Astro plugin

### Result

Tada!

image

Props are always shown first unless something matches better (ex: typing c on that screenshot would instead show the client directives first)

## Testing

Tested manually

## Docs

No docs needed

### DIFF
--- a/packages/language-server/src/plugins/html/features/astro-attributes.ts
+++ b/packages/language-server/src/plugins/html/features/astro-attributes.ts
@@ -1,12 +1,18 @@
 import { newHTMLDataProvider } from 'vscode-html-languageservice';
 
-export const astroAttributes = newHTMLDataProvider('astro-attributes', {
+export const classListAttribute = newHTMLDataProvider('class-list', {
 	version: 1,
 	globalAttributes: [
 		{
 			name: 'class:list',
 			description: 'Utility to provide a list of class',
 		},
+	],
+});
+
+export const astroAttributes = newHTMLDataProvider('astro-attributes', {
+	version: 1,
+	globalAttributes: [
 		{
 			name: 'set:html',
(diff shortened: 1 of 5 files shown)


packages/language-server/src/plugins/html/features/astro-attributes.ts
```diff
@@ -75,6 +81,7 @@ export const astroDirectives = newHTMLDataProvider('astro-directives', {
 		{
 			name: 'client:load',
 			description: 'Start importing the component JS at page load. Hydrate the component when import completes.',
+			valueSet: 'v',
```
Princesseuh, line 84: So, I'm not sure why it's just named `v` instead of something descriptive but this changes the text edit for the completions to just be the value. 

ex: The `client:load` completions now create the following text: `client:load` instead of `client:load=""`

matthewp (approved):
Looks great!

PR #236, opened 2022-03-29
Add a test suite for TypeScript and Astro plugin

Princesseuh:
## Changes

This adds a test suite for our TypeScript and our Astro plugin. This allowed to uncover a few issues throughout all of our plugins, fixes were provided for all issues and will be documented through review comments on this issue

This PR does not include tests for every single feature, it's more so intended as a "starting structure" to build from. Hopefully, having this mean that when additional features will be added, so will tests (the easier it is to write a test, the more likely it is that someone actually does write one!)

Here's a quick coverage report:

**TypeScript**
- [x] HoverInfo
- [x] DocumentSymbols
- [x] SemanticTokens
- [ ] Completions
- [ ] Diagnostics
- [ ] SignatureHelp
- [ ] Definitions
- [ ] Rename

**Astro**
- [x] Folding Ranges
- [x] Completions
- [ ] Definitions

More work will be done on this in further PRs dedicated to specific features

## Testing

Apart from a few fixes, It's tests all the way down

## Docs

No docs needed


packages/language-server/src/plugins/typescript/features/DocumentSymbolsProvider.ts
```diff
@@ -33,7 +34,8 @@ export class DocumentSymbolsProviderImpl implements DocumentSymbolsProvider {
 					Range.create(
 						document.positionAt(document.astroMeta.frontmatter.startOffset as number),
 						document.positionAt(document.astroMeta.frontmatter.endOffset as number)
-					)
+					),
+					document.getURL()
```
Princesseuh, line 38: Previously the symbols for the namespace didn't have the proper URI assigned to them, which meant that in certain case you wouldn't navigate to the proper files when trying to go a document's frontmatter or template

packages/language-server/src/plugins/typescript/features/DocumentSymbolsProvider.ts
```diff
@@ -46,15 +48,18 @@ export class DocumentSymbolsProviderImpl implements DocumentSymbolsProvider {
...
+				document.getURL()
 			)
 		);
 
 		for (let symbol of symbols.splice(1)) {
 			symbol = mapSymbolInformationToOriginal(fragment, symbol);
 
 			if (document.offsetAt(symbol.location.range.end) >= (document.astroMeta.content.firstNonWhitespaceOffset ?? 0)) {
-				symbol.containerName = 'Template';
+				if (symbol.containerName === originalContainerName) {
```
Princesseuh, line 60: Symbols not at the root of the template previously had the wrong containerName. I don't think this caused any issues however

packages/language-server/package.json
```diff
@@ -16,7 +16,7 @@
   "scripts": {
     "build": "tsc",
     "dev": "astro-scripts dev \"src/**/*.ts\"",
-    "test": "cross-env TS_NODE_TRANSPILE_ONLY=true mocha --require ts-node/register \"test/**/*.ts\" --exclude \"test/**/*.d.ts\""
+    "test": "cross-env TS_NODE_TRANSPILE_ONLY=true mocha --timeout 20000 --require ts-node/register \"test/**/*.ts\" --exclude \"test/**/*.d.ts\""
```
Princesseuh, line 19: I do not think that it's necessary our fault that we need this, I think TypeScript is just slow and will get slower as we add more files

packages/language-server/src/plugins/astro/features/CompletionsProvider.ts
```diff
@@ -51,20 +51,19 @@ export class CompletionsProviderImpl {
 			if (frontmatter) items.push(frontmatter);
 		}
 
+		const html = document.html;
 		const offset = document.offsetAt(position);
-		if (!isInsideFrontmatter(document.getText(), offset)) {
+		if (isInComponentStartTag(html, offset)) {
```
Princesseuh, line 56: Without this, we were getting completions for props while being anywhere near a component such as inside the component or inside its ending tag

packages/language-server/src/plugins/typescript/features/SemanticTokenProvider.ts
```diff
@@ -28,7 +28,10 @@ export class SemanticTokensProviderImpl implements SemanticTokensProvider {
 			filePath,
 			{
 				start,
-				length: range ? fragment.offsetAt(fragment.getGeneratedPosition(range.end)) - start : fragment.text.length,
+				length: range
+					? fragment.offsetAt(fragment.getGeneratedPosition(range.end)) - start
+					: // We don't want tokens for things added by astro2tsx
+					  fragment.text.lastIndexOf('export default function (_props:') || fragment.text.length,
```
Princesseuh, line 34: Previously we'd generate tokens for code generated by `astro2tsx`, which I don't think would create issues? But is still technically improper behaviour

packages/language-server/src/plugins/html/features/astro-attributes.ts
```diff
@@ -55,6 +55,17 @@ export const astroAttributes = newHTMLDataProvider('astro-attributes', {
 						},
 					],
 				},
+				{
```
Princesseuh, line 58: Unrelated to tests, I just noticed that those attributes were missing and felt that it was harmless to do it here

packages/language-server/src/plugins/astro/AstroPlugin.ts
```diff
@@ -41,7 +41,7 @@ export class AstroPlugin implements Plugin {
 		return completions;
 	}
 
-	async getFoldingRanges(document: AstroDocument): Promise<FoldingRange[]> {
+	getFoldingRanges(document: AstroDocument): FoldingRange[] {
```
matthewp, line 44: Just a side-note, but we've started always adding a return signature in the main astro repo (informally), and I like to see that happening here as well. Think it helps with bugs.

Princesseuh: Not sure what you mean by a "return signature"? You mean a return type?

PR #237, opened 2022-04-01
Disable CSS completions and hover info in unclosed style tags

Princesseuh:
## Changes

We cannot reliably give CSS completions and hover info for unclosed style tags because if the HTML parser doesn't find an ending tag, it assume that the entire file is part of the tag (which is expected). This leads to completions and hover info providing invalid info (thinking that HTML tags are CSS stuff etc)

I'm not sure if this is really a needed fix because a fair assumption about editor tooling is that usually, if you give it wrong input, it gives you wrong output, but I figured that we had the power to avoid this here and that we might as well. Either way, you get an error from TypeScript telling you to close the tag, so :woman_shrugging: 

## Caveats

I do not think that it's the case because nowadays editors will autoclose tags for you but I wouldn't be surprised if [someone](https://xkcd.com/1172/) relied on this for completions. It wasn't really possible to rely on it in my opinion because the completions didn't really work but, you never know

## Testing

Added tests for this

## Docs

No docs needed

### DIFF
--- a/packages/language-server/src/plugins/css/CSSPlugin.ts
+++ b/packages/language-server/src/plugins/css/CSSPlugin.ts
@@ -57,6 +57,13 @@ export class CSSPlugin implements Plugin {
 
 		const styleTag = this.getStyleTagForPosition(document, position);
 
+		// We technically can return results even for open tags, however, a lot of the info returned is not valid
+		// Since most editors will automatically close the tag before the user start working in them, this shouldn't be a problem
+		if (styleTag && !styleTag.closed) {
+			return null;
+		}
+
+		// If we don't have a style tag at this position, we might be in a style property instead, let's check
 		if (!styleTag) {
 			const attributeContext = getAttributeContextAtPosition(document, position);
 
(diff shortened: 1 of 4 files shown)


packages/language-server/src/plugins/css/CSSPlugin.ts
```diff
@@ -57,6 +57,13 @@ export class CSSPlugin implements Plugin {
...
 
+		// We technically can return results even for open tags, however, a lot of the info returned is not valid
+		// Since most editors will automatically close the tag before the user start working in them, this shouldn't be a problem
+		if (styleTag && !styleTag.closed) {
+			return null;
+		}
+
+		// If we don't have a style tag at this position, we might be in a style property instead, let's check
 		if (!styleTag) {
 			const attributeContext = getAttributeContextAtPosition(document, position);
```
natemoo-re, line 68: Unrelated, but we support completions in the `style` attribute? Cool!

packages/language-server/test/plugins/css/CSSPlugin.test.ts
```diff
@@ -63,6 +63,16 @@ describe('CSS Plugin', () => {
 			expect(globalCompletion, 'Expected completions to contain :global modifier').to.not.be.null;
 		});
 
+		it('should not provide completions for unclosed style tags', () => {
```
natemoo-re, line 66: I love that we have tests now!

natemoo-re (approved):
Amazing work! 🎉

matthewp (approved):
lgtm

PR #238, opened 2022-04-01
Add support for global attribute in style tag to language server

sciencefidelity:
## Changes

Adds support for `<style global>` to the language server.

## Testing

There are currently no tests for specific tag attributes.

## Docs

Documents the use of `global` in IDEs. I'd be happy for suggestions how the proposed wording might be improved.

### DIFF
--- a/packages/language-server/src/plugins/html/features/astro-attributes.ts
+++ b/packages/language-server/src/plugins/html/features/astro-attributes.ts
@@ -70,6 +70,16 @@ export const astroAttributes = newHTMLDataProvider('astro-attributes', {
 						},
 					],
 				},
+				{
+					name: 'global',
+					description: 'Opts-out of automatic CSS scoping, all contents will be globally scoped',
+					references: [
+						{
+							name: 'Astro documentation',
+							url: 'https://docs.astro.build/en/guides/styling/#global-styles',
+						},
+					],
+				},
 			],
 		},
 	],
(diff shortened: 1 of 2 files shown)


packages/language-server/src/plugins/html/features/astro-attributes.ts
```diff
@@ -70,6 +70,16 @@ export const astroAttributes = newHTMLDataProvider('astro-attributes', {
 						},
 					],
 				},
+				{
+					name: 'global',
+					description: 'Opts-out of automatic CSS scoping, all contents will be available globally',
```
Princesseuh, line 75:
```suggestion
					description: 'Opts-out of automatic CSS scoping, all contents will be available globally',
					valueSet: 'v',
```

By adding the valueSet  parameter and setting it to `v`, the completion will be nicer (`global` instead of `global=""`)

Princesseuh:
Good catch! I had a fix for that in https://github.com/withastro/language-tools/pull/236 but your wording is nicer, so I'll remove it from my PR. I just had a little suggestion regarding the completion but everything else looks good to me, nice work!

Princesseuh:
Good catch! I had a fix for that in https://github.com/withastro/language-tools/pull/236 but your wording is nicer, so I'll remove it from my PR. I just had a little suggestion regarding the completion but everything else looks good to me, nice work!

sciencefidelity:
Done, thanks for the suggestion, I didn't know about that.

PR #246, opened 2022-04-07
Ignore "Can't return outside of a function body" diagnostic

Princesseuh:
## Changes

Ignore the "Can't return outside of a function body" diagnostic. While it is technically a valid diagnostic, unfortunately due to how our TSX is formed, the frontmatter is at the top level and thus any return there will create this issue

This wasn't a problem before as we never returned from the frontmatter but now with the addition of SSR, this is something that users might do

## Caveeat

There's still an issue where unconditional returns will result in TypeScript telling us the entire template is unreachable code, which is currently right, as returning a Response does not return the template. However in the future this will change if we were to support returning in the frontmatter while still using the template

Hopefully by then, we'll have a different TSX output :sweat_smile:

## Testing

Tested manually

## Docs

No docs needed

### DIFF
--- a/packages/language-server/test/plugins/typescript/TypeScriptPlugin.test.ts
+++ b/packages/language-server/test/plugins/typescript/TypeScriptPlugin.test.ts
@@ -69,6 +69,30 @@ describe('TypeScript Plugin', () => {
 		});
 	});
 
+	describe('provide diagnostics', async () => {
+		it('return diagnostics', async () => {
+			const { plugin, document } = setup('diagnostics/basic.astro');
+
+			const diagnostics = await plugin.getDiagnostics(document);
+			expect(diagnostics).to.not.be.empty;
+		});
+
+		it('should not provide diagnostics if feature is disabled', async () => {
+			const { plugin, document, configManager } = setup('diagnostics/basic.astro');
+
+			configManager.updateConfig(<any>{
+				typescript: {
+					diagnostics: {
+						enabled: false,
+					},
+				},
+			});
+
+			const diagnostics = await plugin.getDiagnostics(document);
+			expect(diagnostics).to.be.empty;
+		});
+	});
+
 	describe('provide semantic tokens', async () => {
 		it('return semantic tokens for full document', async () => {
 			const { plugin, document } = setup('semanticTokens/frontmatter.astro');
--- a/packages/language-server/test/plugins/typescript/features/DiagnosticsProvider.test.ts
+++ b/packages/language-server/test/plugins/typescript/features/DiagnosticsProvider.test.ts
@@ -1,0 +1,82 @@
+import { expect } from 'chai';
+import { Range } from 'vscode-languageserver-types';
+import { createEnvironment } from '../../../utils';
+import { LanguageServiceManager } from '../../../../src/plugins/typescript/LanguageServiceManager';
+import { DiagnosticsProviderImpl } from '../../../../src/plugins/typescript/features/DiagnosticsProvider';
+
+describe('TypeScript Plugin#DiagnosticsProvider', () => {
+	function setup(filePath: string) {
+		const env = createEnvironment(filePath, 'typescript', 'diagnostics');
+		const languageServiceManager = new LanguageServiceManager(env.docManager, [env.fixturesDir], env.configManager);
+		const provider = new DiagnosticsProviderImpl(languageServiceManager);
+
+		return {
+			...env,
+			provider,
+		};
+	}
+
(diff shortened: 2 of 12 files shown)


matthewp:
We have tests for diagnostics so I would think you could add a test for this, no?

Princesseuh:
> We have tests for diagnostics so I would think you could add a test for this, no?

We do not in fact have tests for diagnostics :sweat_smile: It's one of the few features that is currently not covered by the new test suite

matthewp:
@Princesseuh and I chatted and she's going to add diagnostics tests back, and add a test for this one.

jasikpark:
Where would you run into this diagnostic with valid code? I haven't checked out ssr yet though, so that's probably the answer??

Princesseuh:
> Where would you run into this diagnostic with valid code? I haven't checked out ssr yet though, so that's probably the answer??

Yep! Because in SSR you can return a response directly in the frontmatter, ex:

```astro
---
const cookie = Astro.request.headers.get('cookie');

// if the user is not logged in, redirect them to the login page.
if(!isLoggedIn(cookie)) {
  return Astro.redirect('/login');
}
---
```

JuanM04:
I love that this works!

```astro
---
// user: string | null
const user = Astro.request.headers.get('user')

if(!user) return Astro.redirect('/')
---

<p>{user}</p>
<!-- `user` is of type string!!! -->
```

> There's still an issue where unconditional returns will result in TypeScript telling us the entire template is unreachable code, which is currently right, as returning a Response does not return the template.

I see that as a feature! Nonetheless, when would do something like that?

Princesseuh:
> > There's still an issue where unconditional returns will result in TypeScript telling us the entire template is unreachable code, which is currently right, as returning a Response does not return the template.
> 
> I see that as a feature! Nonetheless, when would do something like that?

It is definitely a feature for now (in addition of being absolutely correct behavior)!

> I love that this works!

The power of TypeScript is amazing when everything is working right :grin:

matthewp (changes requested):
Discussed offline, need tests for diagnostics added back and a test for this one.

packages/language-server/test/utils.ts
```diff
@@ -12,11 +12,11 @@ import { pathToUrl } from '../src/utils';
  * @returns
  */
 export function createEnvironment(filePath: string, baseDir: string, pathPrefix?: string) {
-	const fixtureDir = join(__dirname, 'plugins', baseDir, 'fixtures', pathPrefix ?? '');
+	const fixtureDir = join(__dirname, 'plugins', baseDir, 'fixtures');
```
Princesseuh, line 15: Small change to make tests faster, previously we'd set the root of the workspace in the specific folder of the fixtures (so for instance, `fixtures/diagnostics`) instead of the general folder where fixtures are (`fixtures`).

Since there's no `tsconfig.json` in those folders, the language service would end up going up the directories and loading a bunch of unrelated files which slow down things massively

packages/language-server/test/plugins/typescript/TypeScriptPlugin.test.ts
```diff
@@ -69,6 +69,30 @@ describe('TypeScript Plugin', () => {
...
+			configManager.updateConfig(<any>{
+				typescript: {
+					diagnostics: {
+						enabled: false,
+					},
+				},
+			});
+
+			const diagnostics = await plugin.getDiagnostics(document);
+			expect(diagnostics).to.be.empty;
```
matthewp, line 92: yes! like testing for these features.

packages/language-server/test/plugins/typescript/fixtures/diagnostics/hints.astro
```diff
@@ -0,0 +1,6 @@
+---
+	/** @deprecated */
+const deprecated = "Astro"
```
matthewp, line 3: very clever way to test this

packages/language-server/test/plugins/typescript/features/DiagnosticsProvider.test.ts
```diff
@@ -0,0 +1,82 @@
...
+				tags: [1],
+			},
+		]);
+	});
+
+	it('support return in frontmatter', async () => {
+		const { provider, document } = setup('ssrReturn.astro');
+
+		const diagnostics = await provider.getDiagnostics(document);
+		expect(diagnostics).to.be.empty;
```
matthewp, line 71: this is the one!

matthewp (approved):
Looks great, thanks for adding these tests!

Issue #230, opened 2022-03-24
🐛 BUG: Script tags are not being properly handled as being a different context


Princesseuh:
### Describe the Bug

`script` tags in `.astro` files are a very mythical thing for TypeScript, they're in the same file as other JS/TS code (frontmatter or expressions) but at the same time, they run in a completely different context, you can even inject variables from one context to the other using `define:vars`

This doesn't cause much issues at the moment, not necessarily because we have some cool guards in place but because we just don't support many things inside of script tags. For instance, all diagnostics that are inside a `script` tags are currently ignored. Completions are currently not supported inside of script tags etc. 

At the moment, the only thing that shows this issue is hover info, since it doesn't do anything special for script tags but the reality is that this issue will show up more and more as time go on and we support more features in script tags. 

## Possible solution

An easy solution (provided that TS support this nicely) would be to isolate script tags in a scope but, it should technically already be the case so I'm not sure if that's a real solution.

An harder but that could work nicely solution would be to have another DocumentSnapshot which contain only the script tags and then whenever we detect that something is being asked inside a script tag, instead use that Snapshot. This wouldn't work for semantic tokens however but maybe we could make a special case for that

### Steps to Reproduce

1. Create astro project
2. Add variable to frontmatter
3. Use variable in a script tag even though it's a different context
4. No errors! And hover info will show the type from the frontmatter

PR #252, opened 2022-04-12
Add support for folding ranges for JavaScript

Princesseuh:
## Changes

Add support for folding JavaScript, with this everything is now fold-able (HTML, CSS, JavaScript and frontmatter)!

Support include frontmatter and expressions. Folding inside of script tags is currently not supported (require https://github.com/withastro/language-tools/issues/230)

image

## Testing

Tests added

## Docs

No docs needed (no settings needed as we don't support disabling folding ranges (there's not much point to disabling them..)

### DIFF
--- a/packages/language-server/test/plugins/typescript/features/FoldingRangesProvider.test.ts
+++ b/packages/language-server/test/plugins/typescript/features/FoldingRangesProvider.test.ts
@@ -1,0 +1,49 @@
+import { LanguageServiceManager } from '../../../../src/plugins/typescript/LanguageServiceManager';
+import { createEnvironment } from '../../../utils';
+import { FoldingRangesProviderImpl } from '../../../../src/plugins/typescript/features/FoldingRangesProvider';
+import { expect } from 'chai';
+
+describe('TypeScript Plugin#FoldingRangesProvider', () => {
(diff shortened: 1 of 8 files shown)


packages/language-server/test/plugins/typescript/features/FoldingRangesProvider.test.ts
```diff
@@ -0,0 +1,49 @@
...
+		]);
+	});
+
+	it('does not provide folding ranges for ignored tags', async () => {
+		const { provider, document } = setup('excludedtags.astro');
+
+		const foldingRanges = await provider.getFoldingRanges(document);
+
+		// TypeScript return folding ranges for JSX tags, hence why it still returns something
+		// Those get ignored during normal usage since we prefer leaving that to our HTML plugin
```
Princesseuh, line 39: As noted, this is not an issue because our HTML plugin will return folding ranges for those range first, which editors will prioritize

We could technically exclusively use the TypeScript ones instead of the HTML ones but they're not as good for HTML from my testing, not sure why. I know that in the past, VS Code had TypeScript's folding ranges disabled by default due to issues

matthewp (approved):
Sorry for the late review, this looks great!

PR #253, opened 2022-04-15
Overhaul TypeScript completions

Princesseuh:
## Changes

This PR overhaul the way we did TypeScript completions. Previously, they were only enabled in the frontmatter, which is fine but didn't cover 50% of what an Astro file is. This PR now enables it all throughout the Astro file, which has a lot of implications because notably in some cases we don't really want them (ex: because the HTML or Astro one provide better ones). 

Enabling them in the entire file also means that completions inside JSX expressions are now supported (the only missing part being script tags)

This PR fixes a few long-standing issues with our completions, for instance, kind modifiers were not handled properly, which meant that we were missing information inside of our completions (deprecated stuff now shows as deprecated directly in the completions!). There was also an issue where descriptions were not properly parsed and would show a bunch of `[object Object]`, this is now fixed

Additionally, this PR add supports for completion with code actions in them. Namely, this allows for auto imports. This works not only in the frontmatter but also in the template, for instance typing the following and accepting the completion ([video here](https://cdn.discordapp.com/attachments/878341752799494154/964267101403627520/recording.mp4)):

```
<AstroCompo
```

will result in 

```
---
import AstroComponent from '../components/AstroComponent.astro';
---

<AstroComponent
```

This works for all of our supported frameworks (JSX-based ones, Svelte, Vue, Astro)

## Caveats

Completions are really hard to get perfect in one go, depending on context you want different ordering, different commit characters, there's performance concerns, descriptions and labels to show etc. 

This PR's implementation is definitely not perfect and we'll need to adjust stuff as we receive user feedback on this. However, it's an improvement over the previous state, which is always nice.

## Testing

Tests were added

## Docs

No docs needed


packages/language-server/test/plugins/typescript/features/CompletionsProvider.test.ts
```diff
@@ -0,0 +1,125 @@
...
+		const completions = await provider.getCompletions(document, Position.create(1, 11), {
+			triggerKind: CompletionTriggerKind.Invoked,
+		});
+
+		const item = completions?.items.find((completion) => completion.label === 'MySuperFunction');
+		delete item.data;
+
+		expect(item).to.deep.equal({
+			commitCharacters: ['.', ',', '('],
+			insertText: "import { MySuperFunction$1 } from './imports/definitions';",
```
Princesseuh, line 110: The `$1` here is not an error, TypeScript does in fact position the user's cursor there after this kind of completion. 

I don't find this behavior necessarily good and fun to use, but it is consistent with what you would get inside .ts files

aFuzzyBear:
Erika this is fantastic, !!!

matthewp (approved):
Nice tests!

PR #255, opened 2022-04-18
Misc improvements to Astro plugin completions

Princesseuh:
## Changes

Misc improvements to the completions provided by the Astro plugins to better match the TypeScript plugin after https://github.com/withastro/language-tools/pull/253

Depending on the type of the prop, we now give a different insertText. We also show the type on the right hand side of the completion, which is neat

image

## Testing

No tests needed, we don't have dedicated tests for this but our current tests cover this

## Docs

No docs needed

### DIFF
--- a/.changeset/grumpy-pigs-raise.md
+++ b/.changeset/grumpy-pigs-raise.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/language-server': patch
+---
+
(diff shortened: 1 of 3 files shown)


bholmesdev (approved):
I half-understood the code changes but I trust the test updates here! Mission `complete`

Issue #143, opened 2022-01-01
💡 RFC: Auto-import and auto-fixes


itsMapleLeaf:
### Background & Motivation

Would love to be able to autocomplete and auto-import values in .astro files, like you can with TypeScript. Speeds up dev workflow and lowers tedium by a ton.

### Possible solutions

Implement auto imports and import quick fixes the same way the TypeScript LSP and others do.

### Alternatives considered

None

### Risks, downsides, and/or tradeoffs

Complexity?

### Open Questions

:shrug: 

### Help make it happen!

- [ ] I am willing to submit a PR to implement this change.
- [ ] I am willing to submit a PR to implement this change, but would need some guidance.
- [X] I am not willing to submit a PR to implement this change.

PR #257, opened 2022-04-21
Implement TypeScript code actions

Princesseuh:
## Changes

This adds support for TypeScript code actions! Code actions are what powers features such as "Quick fixes" and "Source actions" in the editor.

image

This works everywhere in the file, frontmatter, jsx expressions, template etc. It also support the Organize Imports and Sort Imports actions

This PR also fix an issue where the first completion that add an import would mistakenly add it on the wrong line (due to the shape of our TSX output)

This fix https://github.com/withastro/language-tools/issues/143

## Caveats

I decided against implementing some code actions that are available in TypeScript, at least for this first implementation. Notably, I did not implement refactors and the fix all code action. 

- Refactors in TypeScript are not too good, even in `ts` files some patterns will break them. Other frameworks either don't implement them or they generate broken code :woman_shrugging: 

- The Fix All action in `ts` files, unlike what one might think, does not actually fix all problems and instead fixes a very limited subset of problems that are not necessarily common problems inside Astro files

- "Remove unused code" isn't supported at the moment. Due to the shape of our TSX output, it tends to mistake some used code for unused and remove code greedily. I'm not sure how to both have this and import completions working at the same time with our current TSX output

## Testing

Added tests for this

## Docs

No docs needed

### DIFF
--- a/packages/language-server/src/plugins/PluginHost.ts
+++ b/packages/language-server/src/plugins/PluginHost.ts
@@ -20,6 +20,8 @@ import {
 	WorkspaceEdit,
 	SymbolInformation,
 	SemanticTokens,
+	CodeActionContext,
+	CodeAction,
 } from 'vscode-languageserver';
 import type { AppCompletionItem, Plugin, LSProvider } from './interfaces';
 import { flatten } from 'lodash';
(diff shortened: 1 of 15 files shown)


packages/language-server/src/plugins/PluginHost.ts
```diff
@@ -146,7 +165,7 @@ export class PluginHost {
 
 		const foldingRanges = flatten(
 			await this.execute<FoldingRange[]>('getFoldingRanges', [document], ExecuteMode.Collect)
-		).filter((completion) => completion != null);
+		);
```
Princesseuh, line 168: This was unnecessary, probably added by accident in a copy paste somewhere

Issue #9, opened 2021-08-11
VSCode: Emmet isn't fully supported


RafidMuhymin:
### What package manager are you using?

yarn

### What operating system are you using?

Windows

### Describe the Bug

Emmet support is incomplete in Astro components. Only basic completions are supported.

For example, if I type 'button' in a astro component, emmet only shows 'button' tag completion. But in JSX and HTML, it shows button, button:d, button:disabled, button:r, button:reset, button:s and button:submit tag completion.

### Steps to Reproduce

Advance completions provide by Emmet should be suppported.

PR #258, opened 2022-04-22
Fix Emmet settings not being loaded, add Emmet support in CSS

Princesseuh:
## Changes

We never told VS Code that we want the Emmet config which meant that we never got VS Code's default settings which allows for extended completions (things like `a:link` and others) :sweat: 

This PR also adds support for Emmet in CSS, giving us Emmet support everywhere now! :partying_face:

image

image

This fix https://github.com/withastro/language-tools/issues/9

## Caveats

The support in CSS has some minors issues due to upstream issues but since those issues seems to shows up in every single VS Code extension that depend on the official VS Code Emmet helper, I figured it was alright

The issues are fairly minor in the grand scheme of things, being things like completions showing when they shouldn't and completions with `+` not working properly (which also happens in the official Emmet extension bundled with VS Code and was marked as wontfix in the helper)

## Testing

Added test that we get Emmet completions in CSS

We can't really test that we load the Emmet config properly from VS Code due to it being something that get passed to us when initializing the server and if VS Code stopped passing things to us, we'd have bigger problems than Emmet not working :sweat_smile:

(We should however test that while initializing the server we actually do populate the config correctly from what's passed to us, but it's a bit outside the scope of this PR as it require building a test suite for the whole initializing part of the server)

## Docs

No docs needed

### DIFF
--- a/packages/language-server/src/plugins/css/CSSPlugin.ts
+++ b/packages/language-server/src/plugins/css/CSSPlugin.ts
@@ -158,10 +158,37 @@ export class CSSPlugin implements Plugin {
 		const cssLang = extractLanguage(cssDocument);
 		const langService = getLanguageService(cssLang);
 
-		const emmetResults: CompletionList = {
-			isIncomplete: true,
+		let emmetResults: CompletionList = {
+			isIncomplete: false,
 			items: [],
 		};
+
+		langService.setCompletionParticipants([
+			{
+				onCssProperty: (context) => {
+					if (context?.propertyName) {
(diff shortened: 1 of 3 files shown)


natemoo-re (approved):
Heck yeah! Great work!

aFuzzyBear:
🥳 Im a big fan of Emmet, the unsung squire who comes into our IDEs and makes life that few keystrokes easier. 
Thank you for helping Emmet get into shape for us 💚💚

PR #259, opened 2022-04-22
Drop Node 12 support

Princesseuh:
## Changes

Removes support for Node 12. What this means VS Code version wise is that we now require a version of VS Code higher or equal to 1.56 (April 2021). Users using a previous version will be stuck on the last version of the extension that supports Node 12 (should be 0.15.0)

VS Code just updated to Node 16 in its latest version, so we still have to support Node 14 for a while :smile:

## Testing

Removed tests for Node 12

## Docs

No docs needed


packages/vscode/package.json
```diff
@@ -19,10 +19,11 @@
     "test": "node ./test/runTest.js"
   },
   "engines": {
-    "vscode": "^1.52.0"
+    "vscode": "^1.56.0"
   },
   "activationEvents": [
-    "onLanguage:astro"
+    "onLanguage:astro",
+    "workspaceContains:astro.config.*"
```
Princesseuh, line 26: Unrelated to the Node 14 changes, I added this so the extension also starts whenever it finds an Astro config file, this is especially convenient for projects who don't have many Astro files

matthewp (approved):
lgtm

PR #262, opened 2022-04-25
fix: handle the case when server settings are not provided

virchau13:
## Changes

If server settings were not provided to the Astro LS, it would crash with:
```
{"jsonrpc":"2.0","id":1,"error":{"code":-32603,"message":"Request initialize failed with message: Cannot read properties of undefined (reading 'astro')"}}
```
This fixes this issue by adding a null check to the server settings.

(I ran into this issue with my configuration of Neovim LSP.)

## Testing

Does this need tests at all?

I'm not exactly sure how to go about testing this, anyway, since `configManager` isn't even used in `LanguageServiceManager`.

## Docs

It's just a bug fix, no docs necessary.

### DIFF
--- a/.changeset/four-laws-occur.md
+++ b/.changeset/four-laws-occur.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/language-server': patch
+---
+
+Fixed bug where nonexistent server settings would result in a crash
--- a/packages/language-server/src/server.ts
+++ b/packages/language-server/src/server.ts
@@ -68,8 +68,14 @@ export function startLanguageServer(connection: vscode.Connection) {
 		}
 
 		// Update language-server config with what the user supplied to us at launch
-		configManager.updateConfig(params.initializationOptions.configuration.astro);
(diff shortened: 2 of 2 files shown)


Princesseuh:
Looks good to me! Thank you for making a PR. 

VS Code never run into this issue since it supplies an empty value if non-existing so I never accounted for it. Good catch!

> Does this need tests at all?

We currently don't test anything related to what is given to us at initialization time, otherwise we could test it there. I don't really mind that there's no tests for this for now, but I'll definitely eventually add a test suite for this (where I'll test this among other things)

> I'm not exactly sure how to go about testing this, anyway, since configManager isn't even used in LanguageServiceManager.

`LanguageServiceManager` is exclusively used for TypeScript-related stuff where effectively, at the moment, we don't use the config at all. However, the config is used in other places, for instance in the different plugins to check if features are enabled

We wouldn't test that there either way though, it'd be in tests for `ConfigManager` or initialization tests as said previously

Thank you again for submitting this change!

Issue #12, opened 2021-08-07
VSCode: Autocomplete for JS, CSS/SCSS inside Astro components


RafidMuhymin:
### What package manager are you using?

yarn

### What operating system are you using?

Windows

### Describe the Bug

Currently, autocomplete doesn't work for JavaScript inside the script tags, CSS, and SCSS inside Astro components.,

### Steps to Reproduce

JavaScript Intellisense now works inside the frontmatter but not inside the script tags. It should work in both of the places. And CSS and SCSS IntelliSense should work inside the style tags.

Princesseuh:
SASS unfortunately cannot be supported at this time because the VSCode's CSS language service only support CSS, LESS & SCSS

Just to update a bit the issue, CSS/SCSS completions should already be working in style tags. The only thing missing there is Emmet

Princesseuh:
https://github.com/withastro/language-tools/pull/258 adds support for Emmet completions inside CSS

As such, the only remaining thing in this issue is completions inside of script tags. That is currently blocked by https://github.com/withastro/language-tools/issues/230

PR #263, opened 2022-04-26
Add support for TypeScript features inside script tags

Princesseuh:
## Changes

This is a first pass at adding support for TypeScript features inside of script tags. JSX does not support script tags so we have to provide support for this manually. Also, TypeScript was really confused about frontmatter context vs script tags.

The way it is done right now is to create a separate snapshot for every script tags in a document so we can do language service stuff on that separately. This is also how CSS works at the moment so it's not a completely unknown situation for us (we already had a mapper exactly for this)

Fix https://github.com/withastro/language-tools/issues/230 and fix https://github.com/withastro/language-tools/issues/12

### Features supported (slightly ordered by priority / how easy to add)

- [x] Hover
- [x] Completions
  - [x] Auto-imports
- [x] Diagnostics
- [x] Code actions
- [x] Folding Ranges
- [x] Signature Helper
- [x] Go to Defintion
- [x] Document Symbols

#### Bonus
- [ ] `define:vars` support
- [ ] `is:inline` support
- [ ] Semantic Token

`define:vars` and `is:inline` are both really important to eventually support. `define:vars` especially is really important to support. However, I am running into difficulties implementing them:

- `define:vars` I am currently unsure how to support efficiently with full TypeScript support. I might wait on the compiler output for this
- `is:inline` script tags requires a completely separate mode of working as they don't operate at all like default Astro ones does. Some sort of basic support for them could be provided using a separate language service (maybe language service from string?)
- Semantic Tokens are, as far as I know, not possible to support due to how they're implemented. I haven't investigated this in depth but my testings so far have been unfruitful

## Caveats

- The TS language service used for the document is re-used for script tags. This is good in most cases but it means that some things that shouldn't be showing are (ex: completion for the Astro global, component imports etc)
- TypeScript settings are used inside script tags instead of JavaScript ones
- Script tags are JavaScript, this means that to be type-checked, `checkJS` needs to be set to `true` or a TS comment needs to be used

## Testing

Tests added

## Docs

No docs needed


packages/language-server/src/server.ts
```diff
@@ -63,8 +64,10 @@ export function startLanguageServer(connection: vscode.Connection) {
 
 		// We don't currently support running the TypeScript and Astro plugin in the browser
 		if (params.initializationOptions.environment !== 'browser') {
-			pluginHost.registerPlugin(new AstroPlugin(documentManager, configManager, workspaceUris));
-			pluginHost.registerPlugin(new TypeScriptPlugin(documentManager, configManager, workspaceUris));
+			const languageServiceManager = new LanguageServiceManager(documentManager, workspaceUris, configManager);
+
+			pluginHost.registerPlugin(new AstroPlugin(languageServiceManager, configManager));
+			pluginHost.registerPlugin(new TypeScriptPlugin(languageServiceManager, configManager));
```
Princesseuh, line 70: Unrelated to the script changes. Previously we created a new LanguageServiceManager for each plugin, this is unnecessary as they can definitely share one. 

This should improve memory consumption and performance

Princesseuh:
This PR is now feature complete. Every single features described work inside script tags. Next steps are to write tests and clean up certain parts and then review time!

matthewp (approved):

PR #265, opened 2022-04-28
Add support for inlay hints

Princesseuh:
## Changes

This adds support for [inlay hints](https://code.visualstudio.com/docs/languages/typescript#_inlay-hints)! This requires upgrading the versions of a few things as inlay hints support was only added in the latest version of the LSP. Those versions are stable, as they've been through a lot of testing (and are already in use by other extensions)

image

## Testing

Tests added

## Docs

No docs needed, inlayHints are enabled using the same configuration they're enabled with in TypeScript files

### DIFF
--- a/.changeset/quiet-olives-appear.md
+++ b/.changeset/quiet-olives-appear.md
@@ -1,0 +1,6 @@
+---
+'@astrojs/language-server': minor
+'astro-vscode': minor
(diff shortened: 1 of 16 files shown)


matthewp (approved):
missing a changeset but otherwise looks good

PR #266, opened 2022-04-29
Overhaul Config Management

Princesseuh:
## Changes

Configuration in language servers can work in two modes: push and pull. The former is what we're currently doing, where the client will push settings to the server at initialization, this mode is deprecated in favor of the later, pull. 

In that mode, the server will pull configurations sections from the client, [this does not make people necessarily happy](https://github.com/microsoft/language-server-protocol/issues/972) and it is in fact, not as easy as implement and use as the previous mode but it allows for per-workspace and per-document settings, which is nice

So this PR migrates to pull mode, additionally this PR includes a refactor to our settings definitions inside the VS Code's extension `package.json` to make sure every settings is there with the proper names. Some settings were removed as they're features we don't support yet anyway. 

We also now respect the user's TS settings inside their editors. Previously we'd either force the default settings or force our own settings. This allow users to control the result of completions and code actions (for instance, if you have `typescript.preferences.quoteStyle` set to `double`, the completion will now always use `"`). 

Some of the default settings are much better than what we forced on, notice how it recommends the aliases paths now, whereas previously it would recommend a relative path:

image

## Testing

Tests updated for new config

## Docs

The new settings are documented by their title and a short description

### DIFF
--- a/packages/language-server/src/plugins/html/features/astro-attributes.ts
+++ b/packages/language-server/src/plugins/html/features/astro-attributes.ts
@@ -42,6 +42,7 @@ export const astroAttributes = newHTMLDataProvider('astro-attributes', {
 		{
 			name: 'is:raw',
 			description: 'Instructs the Astro compiler to treat any children of this element as text',
+			valueSet: 'v',
 			references: [
 				{
 					name: 'Astro reference',
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "eslint-plugin-prettier": "^3.4.0",
     "turbo": "^1.0.0",
     "prettier": "^2.2.1",
-    "typescript": "^4.5.4"
+    "typescript": "^4.6.0"
   },
   "engines": {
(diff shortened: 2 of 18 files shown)


aFuzzyBear:
Nup, no idea, not even the slightest clue.....however it does sound really good!
Week on week, you are crushing it.... 💚💚💚

packages/language-server/src/plugins/html/features/astro-attributes.ts
```diff
@@ -42,6 +42,7 @@ export const astroAttributes = newHTMLDataProvider('astro-attributes', {
 		{
 			name: 'is:raw',
 			description: 'Instructs the Astro compiler to treat any children of this element as text',
+			valueSet: 'v',
```
Princesseuh, line 45: Unrelated to the config change, I just noticed that it had the wrong valueSet while testing things. This is very much harmless so figured I'd do it here

PR #267, opened 2022-05-02
Improve Go to Definition

Princesseuh:
## Changes

Improve our Go To Definition to be a bit smarter and closer to how it behaves inside `.ts` files. Notably, it's now able to link to the exact lines and characters of a definition (in files where it makes sense, so not Astro, Svelte and Vue) instead of linking to the top of the file

I removed the `getGoToDefinitionRefsForImportSpecifier` part because it didn't seems to be relevant anymore. Going to the definition of an alias will lead you to the original file by default :woman_shrugging: 

This also completely remove the definitions provided by the Astro plugin as they were very often just duplicates of the TypeScript ones

## Testing

Added tests for Go To Definition

## Docs

No docs needed


JuanM04 (approved):
I have no clue of what's going on but is those tests pass, LGTM!
Also, a changeset is missing

PR #272, opened 2022-05-06
Add support for props completions from `.d.ts` files, improve performance

Princesseuh:
## Changes

This PR does a few things to the Astro plugin's completions :sweat_smile:

- Add support for getting props from `.d.ts` files, for transpiled `jsx` components (this was discovered through Ben's stream on FrontendHorse where Prismic didn't get completions)
- Improve performance for getting props completions, I added a cache (like the TS's completions) and additionally removed additional unnecessary TS calls as they're super expensive. In local, this made the completion about 150-200ms(!) faster
- Removed `-` as a commit character for the frontend completions. It's not possible for a commit character to not be included in the completion, which meant that users would end up with one too many dash after using the frontmatter completion
- Misc refactors

image

**Note:**
Getting types from complex files is very slow, this is also the case inside `.ts` and `.tsx` files. Prismic especially runs into this a lot, getting props require parsing multiple massive `.d.ts` files. Not much we can do here

## Testing

Tests added and updated

## Docs

No docs needed

### DIFF
--- a/.changeset/itchy-eyes-hear.md
+++ b/.changeset/itchy-eyes-hear.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/language-server': patch
+---
+
+Add support for loading props completions from .d.ts files, improve performance of props completions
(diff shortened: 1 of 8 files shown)


matthewp (approved):

aFuzzyBear:
✨✨✨✨✨

jasikpark:
> Removed - as a commit character for the frontend completions. It's not possible for a commit character to not be included in the completion, which meant that users would end up with one too many dash after using the frontmatter completion

Thanks for fixing that, it's been a problem for a while but I never took the time to create an issue :D

PR #277, opened 2022-05-10
Add support for formatting

Princesseuh:
## Changes

This adds formatting support to the language server. Powered by vscode-html-languageservice and TypeScript. It's pretty 1:1 with the formatting VS Code provide by default inside HTML and JavaScript files, so users used to that will feel right at home, especially since it respect the user's settings for that

The formatting provided is clearly not perfect, as it's mostly intended for HTML and not for JSX (even less so Astro's kind of JSX), however I found the results to be frankly pretty good in most cases (tested using various files for the docs and other examples)

Since it uses Microsoft's API for formatting, it's fairly maintenance-less and it wasn't necessarily hard to implement either

### Caveats

I tested as many pages from real-world usage as I could (our docs, my website etc) and found no case where it generated broken code, however, it can generate broken code inside expressions in some extreme cases. Notably, if a line is extremely long, it can put a newline before an arrow function which is apparently not allowed. There's usually workarounds that are fairly obvious though it's definitely undesirable. If we notice that it happens too much, we could add some sort of protections against that

This is mostly a question of taste, but I find that it sometimes has.. opinions on how to format expressions that I don't always agree with. However, sometimes I find that it creates a better result than Prettier, so maybe it's just taste. There's some (not a lot, however) settings you can configure (both in the Astro extension and in `html.format`) to change the output 

### Why not Prettier?

People who have followed the development of this project might be aware that the plan was to add Prettier to the extension and do formatting that way. This is definitely still a possibility! (We could even have both!)

However, our Prettier plugin and our compiler still needs some work before that can happen, but users are really looking forward to support for formatting so I figured that this could be a nice thing to do in the meantime

I think this could be a nice alternative for users who might want a more configurable alternative to Prettier too

### Why not exclusively use TypeScript's formatting?

TypeScript is able to format JSX files, however its error tolerance is extremely low and as such, most Astro components completely breaks it. For instance, TypeScript will refuse to format any file that has multiple roots, as it's not allowed in JSX. 

A bit more personal but I also don't find the results of it to be very good in the case of JSX. However, it does JS/TS very well so we still use that for the frontmatter and script tags (just like VS Code itself does in `html` files), it works nicely!

## Testing

Tests were added

## Docs

Docs was added

### DIFF
--- a/.changeset/orange-chicken-march.md
+++ b/.changeset/orange-chicken-march.md
@@ -1,0 +1,6 @@
+---
+'@astrojs/language-server': minor
+'astro-vscode': minor
+---
+
+Add support for formatting
--- a/packages/language-server/package.json
+++ b/packages/language-server/package.json
@@ -25,7 +25,7 @@
     "source-map": "^0.7.3",
     "typescript": "~4.6.4",
     "vscode-css-languageservice": "^5.1.13",
-    "vscode-html-languageservice": "^4.2.2",
+    "vscode-html-languageservice": "^4.2.5",
     "vscode-languageserver": "^8.0.0",
     "vscode-languageserver-protocol": "^3.17.0",
     "vscode-languageserver-textdocument": "^1.0.1",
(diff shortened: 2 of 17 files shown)


matthewp:
Thanks for the explanation! I think we should add this and if user feedback is that it's a problem we can always revert.

buster95:
I think code formatting should be part of this vscode extension, if you give this task to prettier we are forced to use prettier

Issue #169, opened 2022-02-28
🐛 BUG: Unable to prop errors importing Vue components


matthewp:
### Describe the Bug

When you import a Vue component, you are not able to get any error messages on missing props.

### Steps to Reproduce

```astro
---
import Counter from '../components/Counter.vue';
---

<Counter />
```

matthewp:
https://www.npmjs.com/package/@vuedx/compiler-tsx

PR #282, opened 2022-05-12
Add support for props completions and type checking for Vue components

Princesseuh:
## Changes

This adds support for getting props from Vue files, similar to what we do for Svelte. Unfortunately, there's about 6 ways you can define props inside a Vue 3 component and it's not very possible for us to support all them. So I decided to focus on supporting the ones that are the most used. 

As such, we only support the following inside `script setup` (so Composition API):

Non-TS way:

```vue
<script setup>
  defineProps({
    name: { type: String, required: true },
  })
</script>
```

TS-way:

```vue
<script setup lang="ts">
  defineProps<{name: string}>()
</script>
```

&

```vue
<script setup lang="ts">
  interface Props {
    name: string
  }

  defineProps<Props>()
</script>
```

This fixes https://github.com/withastro/language-tools/issues/169

## Screenshots

image

image

(JSX props errors are already insanely verbose and for Vue components defined without TypeScript, they're even more verbose, we need to figure out a solution for making props errors clearer at some point, but that's outside the scope of this PR)

For components where props are defined with TS, the error is just like JSX's:

image

## Caveats

Regexes. (@FredKSchott suggested this solution to me, so, all credits to him)

I tried to get it to work using the Vue compiler, but unfortunately the documentation isn't amazing and I have no knowledge of Vue's internals. I then tried using Volar's code-gen package public API but it's also undocumented, there's literally 0 usage of it on GitHub as far as I could find and it always returned empty values so.. I don't know. I also didn't want to involve TypeScript in this as we have to make those snapshots very often and parsing with TypeScript would make it things very slow

To be clear however, even though I do not trust regexes, this works, it's just that it's technically possible to encounter edge cases (ex: you have a defineProps inside a comment) and I do not like that. 

However, this is clearly an improvement over the previous solution which was, well, nothing.

## Testing

Added tests for both supported ways of defining props

## Docs

No docs needed

### DIFF
--- a/packages/vue-language-integration/src/index.ts
+++ b/packages/vue-language-integration/src/index.ts
@@ -1,0 +1,49 @@
+import { parse } from '@vue/compiler-sfc';
+
+export const languageId = 'vue';
+export const extension = '.vue';
+
+export function toTSX(code: string, className: string): string {
+	let result = `export default function ${className}__AstroComponent_(): any {}`;
+
+	// NOTE: As you can expect, using regexes for this is not exactly the most reliable way of doing things
+	// However, I couldn't figure out a way to do it using Vue's compiler, I tried looking at how Volar does it, but I
+	// didn't really understand everything happening there and it seemed to be pretty Volar-specific. I do believe
+	// someone more knowledgable on Vue's internals could figure it out, but since this solution is good enough for most
+	// Vue components (and it's an improvement over, well, nothing), it's alright, I think
+	try {
+		const parsedResult = parse(code);
+
+		if (parsedResult.descriptor.scriptSetup) {
+			const definePropsType = parsedResult.descriptor.scriptSetup.content.match(/defineProps<([\s\S]+)>/m);
+
+			if (definePropsType) {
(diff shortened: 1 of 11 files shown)


packages/vue-language-integration/src/index.ts
```diff
@@ -0,0 +1,49 @@
...
+			if (definePropsType) {
+				result = `
+						${parsedResult.descriptor.scriptSetup.content}
+
+						export default function ${className}__AstroComponent_(_props: ${definePropsType[1]}): any {
+							<div></div>
+						}
+				`;
+			} else {
+				const defineProps = parsedResult.descriptor.scriptSetup.content.match(/defineProps\([\s\S]+\)/m);
```
matthewp, line 29: Since this is at least checking for whitespace I feel okayish about this solution. Definitely better than what we had before.

matthewp:
Definitely not ideal to use regexes, but better than what we had before and we're guilty of abusing regexes in much worse ways, so I think we should merge!

PR #283, opened 2022-05-13
Properly mark optional parameters with a `?`

Princesseuh:
## Changes

This makes it so a `?` is added to optional parameters inside completions, like it does inside `.ts` files. Also added this for props, like in JSX files. Optional props are now sorted below required ones

This makes a really nice result for components with many props, such as `astro-eleventy-img`'s `<Image>`

## Screenshots

image

image

## Testing

Updated tests with new sorting text, added new test for components

## Docs

No docs needed

### DIFF
--- a/.changeset/witty-melons-count.md
+++ b/.changeset/witty-melons-count.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/language-server': patch
+---
+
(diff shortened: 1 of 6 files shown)


matthewp:
lgtm, good tests!

matthewp:
Oh, needs changeset.

Issue #274, opened 2022-05-06
💡 RFC: Use JSX syntax for the template instead of HTML?


Princesseuh:
### Background & Motivation

Our syntax currently assume the template to be HTML, except inside JSX expressions where it turns into, well, JSX

The problem is that we run into a few incompatibility, notably since https://github.com/withastro/compiler/pull/364 it's possible to have JSX comments inside the template and inside elements. Which mean that, in some cases, the user wants HTML template and in some cases JSX ones. It's impossible to know (since they're both valid!) and it cannot be a setting due to how syntax files works

We also run into some issues where JSX template would fix issues, such as https://github.com/withastro/language-tools/issues/7 but there's also other case where it's the reverse such as https://github.com/withastro/language-tools/issues/271

### Possible solutions

I think that overall the JSX syntax is better and closer to Astro. We can probably assume JSX everywhere with perhaps (if that's possible) a few HTML additions such as support for HTML comments. 

As for comments, I am not sure if that's possible but I'd maybe go with HTML comments outside of elements, JSX ones otherwise (including in expressions)?

### Alternatives considered

N/A

### Risks, downsides, and/or tradeoffs

N/A

### Open Questions

Is Astro HTML? Is Astro JSX? No, It's something else 👀

### Help make it happen!

- [X] I am willing to submit a PR to implement this change.
- [ ] I am willing to submit a PR to implement this change, but would need some guidance.
- [ ] I am not willing to submit a PR to implement this change.

Issue #112, opened 2021-11-24
🐛 BUG: Auto-indentaion doesn't work properly


Blinks44:
### Describe the Bug

VScode extension doesn't properly add an indentation in .astro files. I shot a small video - https://yadi.sk/i/GgeCFijpPbicUg

### Steps to Reproduce

See the video above.

Issue #7, opened 2021-08-20
VSCode: JSX {} dynamic operator isn't higlighted


RafidMuhymin:
### What package manager are you using?

yarn

### What operating system are you using?

Windows

### Describe the Bug

The JSX {} dynamic operator isn't higlighted

### Steps to Reproduce

The JSX {} dynamic operator should be higlighted

matthewp:
@RafidMuhymin can you provide an example of how this is highlighted in JSX but is not being highlighted in Astro? In my theme there is no special color provided to either of these. If there's a theme you use where that's not the case please let me know so I can check it out.

RafidMuhymin:
@matthewp Astro:
![image](https://user-images.githubusercontent.com/63650415/144274551-20b4c64c-a456-4d42-a198-bc3afb42e84c.png)

JSX:
![image](https://user-images.githubusercontent.com/63650415/144274828-be9d8577-2eaf-44e1-9bd0-010965b5ed13.png)

RafidMuhymin:
Minimal repro:
```astro
---
// Please view the code in VSCode. Highlighting in GitHub is different from VSCode
---

<!DOCTYPE html>
<html>
  <body>
    <!-- The JSX {} operator isn't highlighted property in the following lines -->
    <h1>Hello {src}!</h1>
    <img {src} />

    <!-- The inner JSX {} operator in the following line is highlighted properly -->
    {(<img src={src} />)}
  </body>
</html>
```

Princesseuh:
This was fixed, however something to note is that you very rarely see native highlighting for brackets these days due to VS Code's bracket colorization. So to see the updated colors, you'd have to disable that

Issue #285, opened 2022-05-16
🐛 BUG: `is:raw` is not supported


Princesseuh:
### Describe the Bug

Currently, the `is:raw` directive is completely ignored. We should do the following:

- In our syntax file, `is:raw` elements should be highlighted as plain text
- Every feature should be disabled inside `is:raw` elements (no completions, no diagnostics etc)

### Steps to Reproduce

1. Create an element with `is:raw`
2. Element is treated as a normal HTML elements

PR #286, opened 2022-05-16
Syntax and language configuration fixes and improvements

Princesseuh:
## Changes

This fixes a few issues with our syntax file and our language configuration file and also adds some missing features:

- Added support for `is:raw` on script tags, style tags, components and other elements
- Added support for JSON script tags using `application/ld+json` (I think this makes us one of the rare ones to support this, I thought it was neat and fairly easy so, why not)
- Added support for script tags with unknown types (treated as unknown content)
- Added template literals as surrounding pair for autoclose support
- Added support for auto closing `{/* */}`
- Added support for automatically indenting certain patterns (HTML, brackets etc)
- Added better support for identifying words in the file (the frontmatter dashes are no longer included, yay!)
- Fixed expressions brackets not using JSX's coloring
- Fixed leaving the frontmatter into the template making an indent

Fix https://github.com/withastro/language-tools/issues/7
Fix https://github.com/withastro/language-tools/issues/112 however, auto indent is super hard to get right and very flaky. Even in HTML and TSX files it doesn't always works very nicely. I ended up mostly copying the TSX's auto indent rules with a few tweaks from the HTML ones
Fix https://github.com/withastro/language-tools/issues/274 - It doesn't really fix it but making this PR made me realize that it's not a really possible thing to do the way I thought we could do it when I wrote that issue and that things worked differently than I thought, so it's no longer relevant

Half of https://github.com/withastro/language-tools/issues/285

I tried to do https://github.com/withastro/language-tools/issues/271, but I couldn't figure it out. Including the `html:comment` pattern inside expressions worked in some cases but didn't in others and I don't really get why. I then tried to inject HTML comments inside the JSX grammar and ended up with nothing more than an headache. If someone has a solution, I'll gladly take it ha!

## Testing

Added a test suite powered by [vscode-tmgrammar-test](https://github.com/PanAeon/vscode-tmgrammar-test) with instructions on how to use and when to update the snapshots

## Docs

No docs needed?


packages/vscode/languages/astro-language-configuration.json
```diff
@@ -3,7 +3,6 @@
     "blockComment": ["<!--", "-->"]
   },
   "brackets": [
-    ["---", "---"],
```
Princesseuh, line 6: Removed the frontmatter bracket because it causes an indent bump into the template (since the entry and the exit of the bracket are the same characters). If we want to provide an indent into the frontmatter, we should use entry rules

packages/vscode/languages/astro-language-configuration.json
```diff
@@ -5,8 +5,7 @@
   "brackets": [
     ["<!--", "-->"],
     ["<", ">"],
-    ["{", "}"],
-    ["(", ")"]
```
Princesseuh, line 10: There's no cases where you want parenthesizes to count as brackets in the `source.astro` scope, in context where they do matter (ex: `source.ts`, `source.js`, `source.tsx`) the languages used there will add them as brackets

packages/vscode/languages/astro-language-configuration.json
```diff
@@ -16,13 +14,15 @@
     { "open": "'", "close": "'" },
     { "open": "\"", "close": "\"" },
     { "open": "`", "close": "`" },
-    { "open": "<!--", "close": "-->", "notIn": ["comment", "string"] },
+    { "open": "{/*", "close": " */}", "notIn": ["comment", "string"] },
+    { "open": "<!--", "close": " -->", "notIn": ["comment", "string"] },
```
natemoo-re, line 18: Is it intentional that the close has a leading space?

packages/vscode/languages/astro-language-configuration.json
```diff
@@ -33,5 +33,36 @@
...
+      },
+      "action": {
+        "indent": "indent"
+      }
+    }
+  ],
+  "indentationRules": {
+    "increaseIndentPattern": "<(?!\\?|(?:area|base|br|col|frame|hr|html|img|input|keygen|link|menuitem|meta|param|source|track|wbr)\\b|[^>]*\\/>)([-_\\.A-Za-z0-9]+)(?=\\s|>)\\b[^>]*>(?!.*<\\/\\1>)|<!--(?!.*-->)|\\{[^}\"']*$",
+    "decreaseIndentPattern": "^\\s*(<\\/(?!html)[-_\\.A-Za-z0-9]+\\b[^>]*>|-->|\\})"
   }
```
natemoo-re, line 67: Wow, this will be great! Nice job ☺️

packages/vscode/syntaxes/astro.tmLanguage.json
```diff
@@ -68,25 +68,64 @@
...
+            },
+            "3": {
+              "name": "punctuation.definition.tag.end.astro"
+            }
+          },
+          "contentName": "source.unknown"
+        },
+        {
+          "name": "meta.tag.component.astro",
+          "begin": "(</?)([$A-Z_][^/?!\\s<>]*|[^/?!\\s<>.]+\\.[^/?!\\s<>]+)\\b",
```
natemoo-re, line 109: This looks unchanged, but just want to confirm that `namespace.component` is properly highlighted as a component?

packages/vscode/syntaxes/astro.tmLanguage.json
```diff
@@ -131,6 +170,61 @@
         }
       },
       "patterns": [
+        {
+          "comment": "Treat script tag with application/ld+json as JSON. This needs to be higher than is:raw since it's a very possible situation to have is:raw on a JSON script tag",
+          "begin": "\\G(?=\\s*[^>]*?type\\s*=\\s*(['\"]|)(?i:application/ld\\+json)\\1)",
+          "end": "(?=</|/>)",
```
natemoo-re, line 176: Love to see it!

packages/vscode/syntaxes/astro.tmLanguage.json
```diff
@@ -131,6 +170,61 @@
...
+              "begin": "(?<=>)(?!</)",
+              "end": "(?=</)",
+              "contentName": "source.json",
+              "patterns": [{ "include": "source.json" }]
+            },
+            { "include": "#html:tag-attributes" }
+          ]
+        },
+        {
+          "comment": "Treat script tag with is:raw as plain text",
```
natemoo-re, line 188: Curious about the highlighting here... I don't think `is:raw` has any effect on `script` or `style` (braces are already ignored), so do we really want highlighting to change to plaintext? I think everything will still be parsed and bundled.

packages/vscode/test/grammar/test.mjs
```diff
@@ -0,0 +1,55 @@
...
+	const code = await promisifySpawn(process.platform === 'win32' ? 'yarn.cmd' : 'yarn', args, {
+		stdio: 'inherit',
+	});
+
+	if (code > 0) {
+		process.exit(code);
+	}
+}
+
+snapShotTest();
```
natemoo-re, line 55: These tests look awesome!

natemoo-re (approved):
Looks fantastic! None of these comments are blocking, just curious to hear your thoughts.

packages/vscode/languages/astro-language-configuration.json
Princesseuh: Yes! For comments it makes for a nicer experience since people usually add a space at the end. This is actually a consistency fix, we already had the space for JS comments but not HTML ones

packages/vscode/syntaxes/astro.tmLanguage.json
Princesseuh: Ah, I might have misunderstood what `is:raw` does on those tags then! 

I thought it would act like a more intense `is:inline` basically. Especially for script tags, I felt like having a weird type + `is:raw` could be a "common" case

packages/vscode/syntaxes/astro.tmLanguage.json
Princesseuh: Yep, what gives it the special highlighting is actually the `support.class.component.astro` scope. Since it makes it highlighted like a class (in TSX files, it's `support.class.component.tsx`)

packages/vscode/syntaxes/astro.tmLanguage.json
Princesseuh: Decided to remove `is:raw` as plaintext for style tags and script tags after confirming that it doesn't do anything in Astro

PR #288, opened 2022-05-18
Disable labelDetails for the time being as they're buggy

Princesseuh:
## Changes

labelDetails are a new feature of LSP 3.17 that I sneaked in while working on completions a while ago. Since it's just a question of setting a property to a string, I felt that it was harmless and would be a quick nice and easy win

However, it seems like there's some issues, first of all, it seems like when we resolve the completion, we overwrite the label details somehow which leads us to having to set it again, which iI don't believe is intended. Additionally, some completions are missing their label details when building the list, which makes them appear only when we resolve them which isn't desirable

I'll revisit the way we build labels and descriptions in general inside completions in the future as it's not the only part that has inconsistency with TS's and it would be nice to simplify the code

## Testing

No tests changes

## Docs

No docs needed, this wasn't documented anywhere


natemoo-re (approved):
✅

PR #295, opened 2022-05-21
[WIP] Refactor diagnostics, cleanup ignored ones

Princesseuh:
## Changes

- Small refactor to make things a bit cleaner
- Ignore diagnostics after the last line of the file, as they're usually more confusing than helpful
- Updated ignored diagnostics codes:
  - Removed ignoring "Cannot import `.ts` path" one except in script tags where it's still relevant. Since `.js` extensions are [currently not supported](https://github.com/withastro/astro/issues/3415). I made it so it recommends to use no extension instead
  - Made the "spread expected" one more precise (now only check for spread and inside the template) as it was silencing A LOT of valid syntax errors, which caused a lot of confusion among users
  - Removed ignoring "Cannot import JSON modules without the JSON module flag" and instead made it so well, we pass the required flag. That way users get [native support for JSON modules (completions, typechecking etc)](https://user-images.githubusercontent.com/3019731/169626276-fc7380a4-b515-412e-9d00-db7318ff148f.png)
  - Removed both codes ignored by `isNoJSXImplicitRuntimeWarning`. The first error is admittedly, an annoying one, but it's a valid error. The second one is a valid error in all cases

## Testing

WIP

## Docs

No docs required

### DIFF
--- a/packages/language-server/src/plugins/typescript/language-service.ts
+++ b/packages/language-server/src/plugins/typescript/language-service.ts
@@ -294,6 +294,7 @@ async function createLanguageService(
 			noEmit: true,
 			declaration: false,
 
+			resolveJsonModule: true,
 			allowNonTsExtensions: true,
 			allowJs: true,
 			jsx: ts.JsxEmit.Preserve,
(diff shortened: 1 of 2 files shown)


packages/language-server/src/plugins/typescript/language-service.ts
```diff
@@ -294,6 +294,7 @@ async function createLanguageService(
 			noEmit: true,
 			declaration: false,
 
+			resolveJsonModule: true,
```
Princesseuh, line 297: This is part of our starter `tsconfig.json` and also is in our TypeScript docs. So a lot of users already had this, luckily. Now everyone will have it :smile:

matthewp (approved):
lgtm

PR #298, opened 2022-05-24
Update VS Code README and marketplace page

Princesseuh:
## Changes

Update our README with more updated info on supported features and some commonly missed info (ex: how to change the formatting settings)

Also updated our marketplace page with some tags and proper categories to improve discovery

## Testing

N/A

## Docs

I guess it's technically docs

### DIFF
--- a/packages/vscode/package.json
+++ b/packages/vscode/package.json
@@ -46,12 +46,12 @@
   "dependencies": {
     "@astrojs/language-server": "0.19.0",
     "@astrojs/ts-plugin": "0.2.1",
-    "vscode-languageclient": "^8.0.1",
-    "vscode-tmgrammar-test": "^0.1.1"
+    "vscode-languageclient": "^8.0.1"
   },
   "devDependencies": {
     "@types/vscode": "^1.67.0",
     "@vscode/test-electron": "^2.1.3",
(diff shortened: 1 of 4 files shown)


packages/vscode/package.json
```diff
@@ -215,13 +228,15 @@
         },
         "astro.format.indentFrontmatter": {
           "type": "boolean",
-          "default": true,
-          "title": "Formatting: Indent the frontmatter by one level of indentation"
+          "default": false,
```
Princesseuh, line 231: It was decided in #ecosystem on Discord that this one should be turned off by default

packages/vscode/package.json
```diff
@@ -2,6 +2,21 @@
...
+  "qna": false,
+  "keywords": [
+    "astro",
+    "node",
+    "typescript",
+    "javascript",
+    "vue",
+    "svelte",
+    "Static site generator"
+  ],
```
Princesseuh, line 19: All the changes here are inspired by what other extensions do. It might seems weird for the Astro extension to be in the "Formatters" category for example, but since it formats Astro files, it's considered as such

natemoo-re (approved):
Looks great!

PR #299, opened 2022-05-24
[WIP] Add command to open dev server in a side panel

Princesseuh:
## Changes

This adds a command to open the currently running dev server / a new one if there's none in a side panel in the editor. This is powered by VS Code's simple-browser (HMR is supported, clicking between the pages work etc). It loads the user's config using Proload to get the user's specified port if there's any

Some users were actually already doing this by using VS Code's Simple Browser command, however they had to start the dev server separately and type in the address manually

image

## Testing

Tested manually on Windows and Linux 

Our test setup for VS Code is currently not setup in a way that allows this to be tested as it doesn't load a project. Additionally, in a lot of way it would feel like testing things outside our world (is Proload loading the config? Does VS Code's browser work?)

## Docs

No docs needed (added commands are documented in the "Feature contributions" tab on the extension page inside VS Code automatically)

### DIFF
--- a/packages/vscode/src/index.ts
+++ b/packages/vscode/src/index.ts
@@ -7,9 +7,6 @@ import {
 	Position,
 	TextDocumentChangeEvent,
 	ViewColumn,
-	RelativePattern,
-	WorkspaceFolder,
-	Uri,
 } from 'vscode';
 import {
 	LanguageClient,
@@ -136,16 +133,18 @@ export async function activate(context: ExtensionContext) {
 			 * Return a currently running dev server's URL
 			 */
 			async function getDevServerUrl() {
+				const configuredPort = workspace.getConfiguration('astro.devServer').get<number>('port') ?? 3000;
+
 				if (window.activeTextEditor) {
 					const workspaceFolder = workspace.getWorkspaceFolder(window.activeTextEditor?.document.uri);
 
 					if (workspaceFolder) {
-						return await getCurrentServer(workspaceFolder);
+						return await getCurrentServer(workspaceFolder, configuredPort);
 					}
 				} else if (workspace.workspaceFolders && workspace.workspaceFolders.length === 1) {
-					return await getCurrentServer(workspace.workspaceFolders[0]);
+					return await getCurrentServer(workspace.workspaceFolders[0], configuredPort);
 				} else {
-					return await getCurrentServer();
+					return await getCurrentServer(undefined, configuredPort);
 				}
 			}
 		})
--- a/packages/vscode/src/index.ts
+++ b/packages/vscode/src/index.ts
@@ -1,4 +1,16 @@
-import { window, commands, workspace, ExtensionContext, TextDocument, Position, TextDocumentChangeEvent } from 'vscode';
+import {
+	window,
+	commands,
+	workspace,
+	ExtensionContext,
+	TextDocument,
+	Position,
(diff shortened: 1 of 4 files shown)


packages/vscode/src/utils.ts
```diff
@@ -0,0 +1,62 @@
+import * as http from 'http';
+import { WorkspaceFolder } from 'vscode';
+import load from '@proload/core';
```
natemoo-re, line 3: This will match Astro's built-in support for `.ts` (etc) files! Don't think you need to add `@proload/plugin-typescript` since `@proload/plugin-tsm` handles this.

```suggestion
import load from '@proload/core';
import loadTypeScript from '@proload/plugin-tsm';

load.use([loadTypeScript]);
```

packages/vscode/src/index.ts
```diff
@@ -91,6 +104,50 @@ export async function activate(context: ExtensionContext) {
...
+			}
+
+			commands.executeCommand('simpleBrowser.api.open', url, {
+				viewColumn: ViewColumn.Beside,
+			});
+
+			/**
+			 * Return a currently running dev server's URL
+			 */
+			async function getDevServerUrl() {
```
natemoo-re, line 138: Would be super cool if the page in the browser synced with the current page you're editing... too wild?

natemoo-re (commented):
Excited for this!

packages/vscode/src/utils.ts
Princesseuh: So, I tried to use `load.use` and ran into a strange error about the function not existing. I marked this PR as WIP because I wanted to ask you about this 😅

packages/vscode/src/index.ts
Princesseuh: That'd be super cool! We have all the information needed to do that, I think 🤔

packages/vscode/src/index.ts
```diff
@@ -91,6 +104,50 @@ export async function activate(context: ExtensionContext) {
...
+		}),
+		commands.registerCommand('astro.openDevServerPanel', async () => {
+			// Let's first try to find if we have a dev server currently running already
+			let url = await getDevServerUrl();
+
+			// If not, let's run one
+			if (!url) {
+				const terminal = window.createTerminal({ name: 'astro:dev' });
+				terminal.show();
+				terminal.sendText('node ./node_modules/astro/astro.js dev');
```
okikio, line 116: Is the intention to have users already include Astro in their package.json files

packages/vscode/src/index.ts
okikio: ^ Would this work similarly to how the [astro-repl](https://astro.build/play) does?

packages/vscode/src/index.ts
Princesseuh: Yes, this assumes that Astro and its requisites are installed and ready to be started

For users who don't have Astro installed, the language server send a notification on startup about how the experience might be degraded due to that, so it doesn't bother me too much to assume that users have Astro ready to go (this is also inline with how other language servers tend to work)

Princesseuh:
Closing due to inactivity, the code ended up being weirdly complicated anyway. Astro does not use proload anymore to load the Astro config either, so maybe we need to find another solution (could be great if Astro itself gave the port to us)

PR #300, opened 2022-05-24
Add support for Linked Editing

Princesseuh:
## Changes

Simple PR adding support for linked editing inside HTML, it's powered by our usual HTML service and as such, matches VS Code's behavior inside HTML files, down to its downfalls (ex: if you type too fast before the linked range appear, it doesn't work)

This feature is unfortunately disabled by default inside VS Code, not sure why. It's also not very pretty, but it's really useful when you have it enabled!

## Testing

Added a test

## Docs

No docs needed

### DIFF
--- a/.changeset/tiny-radios-train.md
+++ b/.changeset/tiny-radios-train.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/language-server': patch
+---
+
+Add support for giving linked editing ranges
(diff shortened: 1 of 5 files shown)


natemoo-re (commented):
Awesome addition!! ❤️

natemoo-re (approved):

PR #301, opened 2022-05-24
Add support for Svelte components documentation

Princesseuh:
## Changes

Svelte components can be documented using an HTML comment inside the template with a `@component` directive. However, due to how we form our TSX output, the resulting JSDoc comment from svelte2tsx would end up in the wrong place and it wouldn't show when users hovered Svelte components. This PR fix this

image

## Testing

Added a test

## Docs

No docs needed


natemoo-re (approved):
Works for me! I'm sure folks will be thrilled to have this!

PR #303, opened 2022-05-25
Add support for highlighting expressions inside Markdown files

Princesseuh:
## Changes

Adds support for highlighting {} blocks inside Markdown files

## Caveeats

Without a custom file format, it's impossible for this to not have side-effect, as in, if you have the extension installed, even in non-Astro projects, you'll get {} highlighted. There's no way to fix this as grammars don't respect `activationEvents`

## Testing

We can't really test this since it's an injection into another grammar

## Docs

No docs needed


Princesseuh:
Closing this as the side effects are undesirable. What we should probably do is create an "Astro Markdown syntax" that has all the Markdown + this + other Astro stuff (ex: highlighting for `setup: |`)

PR #306, opened 2022-05-26
Add error handling for Svelte/Vue components

Princesseuh:
## Changes

In the language server, we have the possibility to know when a Svelte or a Vue component have errors that makes them not-parseable. For Vue, the sfc-compiler will return errors and for Svelte, `svelte2tsx` will throw

This allow us to return a different TSX output whenever there's an error which in turn, allow us to create an invalid JSX component that will make it so an error appear inside the Astro file directly, and with a little string replace, we can give a more helpful error message :smile:

Unfortunately, we can't show the actual error as we can't get it from the diagnostic (the diagnostic only see that x function is not valid, it doesn't actually know the file it comes from). There's probably ways to work around this, but that'll be a future side goal

image

This is unfortunately not possible in the same way for JSX (React, Preact and Solid) components, as long as they're in the correct format, even with syntax errors they'll be considered potential valid JSX components

## Testing

Added a test

## Docs

No docs needed


packages/vue-language-integration/src/index.ts
```diff
@@ -4,7 +4,7 @@ export const languageId = 'vue';
 export const extension = '.vue';
 
 export function toTSX(code: string, className: string): string {
-	let result = `export default function ${className}__AstroComponent_(): any {}`;
+	let result = `export default function ${className}__AstroComponent_(_props: Record<string, any>): any {}`;
```
Princesseuh, line 7: Unrelated to the changes, this fix the language server complaining about props even in cases where it couldn't tell what props were used. Now, when we can't parse the props, we'll just assume that everything is okay

natemoo-re (approved):
Clever! Looks great.

Issue #318, opened 2022-06-12
🐛 BUG: VSCode doesn't complete the frontmatter if three dashes are already typed


RafidMuhymin:
### Describe the Bug

If I have already typed three dashes `---`, the completion snippet shows up but it doesn't complete the frontmatter for me. The completion snippet should work just fine even if I have typed all the three dashes.

### Steps to Reproduce

1. Go to a `.astro` file
2. At the top of the file, type three dashes `---`
3. Press `tab` to use the snippet shown in the suggestion

axelf:
This is still not working. If i type 3 dashes and press tab this is the result (the 3 dashes are repeated)

`------`

`---`

Expected result is this:

`---`

`---`

PR #320, opened 2022-06-20
Fix frontmatter completion not working after three dashes

Princesseuh:
## Changes

Fixes the frontmatter completion not working when the user had already typed three dashes. The reason it didn't work was because after three dashes, the frontmatter was considered in an "open" state which meant that we instead provided a completion that closed the frontmatter, however since the line already had three dashes, it just did.. nothing

Fix https://github.com/withastro/language-tools/issues/318

## Testing

Added tests for this and other situations

## Docs

No docs needed


natemoo-re (approved):
Ahh love this! Thank you for fixing this.

matthewp (approved):
💜

PR #322, opened 2022-06-21
Add support for TypeScript inside hoisted script tags

Princesseuh:
## Changes

Editor tooling part of https://github.com/withastro/astro/pull/3665. 

This was not implemented in a retro-compatible way, which mean that users using outdated Astro versions might run into false-positive where TS doesn't properly report that TS is not supported inside script tags, I think that's okay since it shouldn't happen too often (and we're <1.0 still)

## Testing

Added a test for this, updated tests that needed changes

## Docs

No docs needed on our side, planning PR / collab for docs update however

### DIFF
--- a/packages/language-server/src/plugins/typescript/utils.ts
+++ b/packages/language-server/src/plugins/typescript/utils.ts
@@ -11,7 +11,7 @@ import {
 	SemanticTokenTypes,
 	SemanticTokensLegend,
 } from 'vscode-languageserver';
-import { AstroDocument, mapRangeToOriginal } from '../../core/documents';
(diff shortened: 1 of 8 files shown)


packages/language-server/src/plugins/typescript/utils.ts
```diff
@@ -397,6 +397,18 @@ export function ensureRealFilePath(filePath: string) {
 	}
 }
 
+/**
+ * Return if a script tag is TypeScript or JavaScript
+ */
+export function getScriptTagLanguage(scriptTag: TagInformation): 'js' | 'ts' {
+	// Using any kind of attributes on the script tag will disable hoisting, so we can just check if there's any
```
matthewp, line 404: nice comment!

PR #327, opened 2022-07-06
Add config setting to support arbitrary attributes

Princesseuh:
## Changes

This adds a setting `astro.typescript.allowArbitraryAttributes` to enable support for non-standard attributes. This is part of the puzzle to support frameworks that want users to add attributes to HTML elements such as Alpine and petite-vue

Additionally, while testing this I realized that there was an issue where for clients that don't support `workspace/configuration`, every single `getConfig` call that used a path instead of a simple string (ex: `astro.format` instead of `astro`) would return `undefined` so I fixed that

## Testing

Added a test

## Docs

The added setting has a description and a title to indicate what it does and when it can be useful

### DIFF
--- a/packages/language-server/src/core/config/ConfigManager.ts
+++ b/packages/language-server/src/core/config/ConfigManager.ts
@@ -51,7 +51,7 @@ type DeepPartial<T> = T extends Record<string, unknown>
  * For more info on this, see the [internal docs](../../../../../docs/internal/language-server/config.md)
  */
 export class ConfigManager {
-	public globalConfig: Record<string, any> = { astro: defaultLSConfig };
+	private globalConfig: Record<string, any> = { astro: defaultLSConfig };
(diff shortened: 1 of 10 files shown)


packages/language-server/src/core/config/ConfigManager.ts
```diff
@@ -1,4 +1,4 @@
-import { merge } from 'lodash';
+import { merge, get } from 'lodash';
```
matthewp, line 1: i thought you had removed lodash previously?

matthewp (approved):

packages/language-server/src/core/config/ConfigManager.ts
Princesseuh: Not yet, it's still used here and inside the pluginHost. I am still planning to get it removed soonish though

PR #330, opened 2022-07-07
Update default settings behaviour when no settings are provided

Princesseuh:
## Changes

Settings now work on an "opt-out" basis, where if a specific settings isn't provided, it'll fallback to its default value (which for features, means enabled)

The reason for this change is that in clients where you use the language server directly, for instance Neovim, the language server will in most cases be started with no settings at all and it's annoying for users to need to enable features all explicitly. For VS Code, this does not change anything because it always ship with all the settings at their default values

## Testing

This is covered already by our tests (if this broke feature checks, pretty much every test would fail). Additionally, we already have tests that test when features are disabled

## Docs

No docs needed. I'll inform people in Discord about the change for those who asked about it

### DIFF
--- a/.changeset/little-chefs-enjoy.md
+++ b/.changeset/little-chefs-enjoy.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/language-server': patch
+---
+
+Updated behaviour when no settings are provided. All features are now considered enabled by default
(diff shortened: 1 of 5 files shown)


matthewp:
Yay, Neovim improvement!

PR #331, opened 2022-07-07
Update component check so completions work inside namespaced components

Princesseuh:
## Changes

Updated component check so we can get completions after the dot in `<thing.>`. This allows notably to get completion for components under a namespace and also for the `Content` component when importing Markdown (either through a dynamic imports or through `Astro.glob`)

image

Additionally, this made it so I could remove some unused helper functions, codebase is now smaller!

## Testing

Added a test

## Docs

No docs needed

### DIFF
--- a/.changeset/violet-deers-warn.md
+++ b/.changeset/violet-deers-warn.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/language-server': patch
+---
+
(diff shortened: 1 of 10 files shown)


Princesseuh:
Err, test is failling on Windows. 

I thought this would be a safe way to test it cross-platform but Windows in CI is always there to kill my joy, will make it simpler 

(return results are different on Windows due to, as far as I know, difference in behaviour in TypeScript itself)

PR #332, opened 2022-07-08
Load language integrations from the user project

Princesseuh:
## Changes

Editor tooling part of https://github.com/withastro/astro/pull/3864

This deprecate both `@astrojs/svelte-language-integration` and `@astrojs/vue-language-integration`

This makes the language-server about 600 files lighter and about 10mb~ less. Since we need those integrations in dev, this only affects users and not developers, but that's okay

## Testing

Our current tests already cover importing and doing stuff with Vue and Svelte components. Adding the integrations as dev dependencies is enough to have them inside fixtures

Not sure if anything more is needed here, testing the package imports would be testing that Node's require work

## Docs

Will tell people in my announce message. I'll monitor #support-threads after to see if we need to document this more heavily or not, it shouldn't technically affect many people (if you're using Svelte, you're likely to have the Svelte integration, I think?)


.eslintignore
```diff
@@ -1,5 +1,6 @@
-.github
```
Princesseuh, line 1: This is unrelated to this specific PR, just something I missed in https://github.com/withastro/language-tools/pull/334

matthewp:
Just needs a changeset.

PR #335, opened 2022-07-11
[WIP] Use the compiler's TSX output

Princesseuh:
## Changes

Previously, we used dubious regexes to convert Astro files to TSX in order to get TypeScript working. This PR changes this so we instead a TSX output and its sourcemaps provided by the Astro compiler 

## Testing

WIP

## Docs

WIP


packages/vscode/scripts/build-browser.js
```diff
@@ -36,6 +36,7 @@ require('esbuild').build({
 	platform: 'browser',
 	format: 'iife',
 	tsconfig: './tsconfig.json',
+	external: ['synckit'],
```
Princesseuh, line 39: `synckit` can't run in the browser! Haven't tried with polyfills, but I doubt it will work even then. Something to investigate when we'll add support for TypeScript in the browser

packages/vscode/scripts/build-node.js
```diff
@@ -15,7 +15,7 @@ require('esbuild')
 		bundle: true,
 		sourcemap: isDev ? true : false,
 		outdir: './dist/node',
-		external: ['vscode'],
+		external: ['vscode', './TSXWorker'],
```
Princesseuh, line 18: TODO: Package this correctly, the worker needs to be a separate file so it's not included in the bundle.

packages/vscode/scripts/build-browser.js
jasikpark: can the browser just use async / shared webworkers instead?

.changeset/tasty-schools-end.md
```diff
@@ -0,0 +1,6 @@
+---
+'@astrojs/language-server': minor
+'astro-vscode': minor
+---
+
+Migrate the language-server to use a new TSX output using the Astro compiler. This should make things such as autocomplete and hover information much more accurate, in additiin to bringing support for numerous Astro features that were previously not working (such as support for the shorthand syntax for props, support for `is:raw` and more!)
```
natemoo-re, line 6:
```suggestion
Migrate the language-server to use a new TSX output using the Astro compiler. This should make things such as autocomplete and hover information much more accurate, in addition to bringing support for numerous Astro features that were previously not working (such as support for the shorthand syntax for props, support for `is:raw` and more!)
```

packages/language-server/src/plugins/typescript/astro2tsx.ts
```diff
@@ -1,111 +1,8 @@
...
-	const astroGlobalConstDef = `
-	/**
-	 * Astro global available in all contexts in .astro files
-	 *
-	 * [Astro documentation](https://docs.astro.build/reference/api-reference/#astro-global)
-	 */
-	declare const Astro: Readonly<AstroGlobal>;
-	`;
+export default function (content: string, fileName: string): TSXResult {
+	const tsx = convertToTSX(content, { sourcefile: fileName });
```
natemoo-re, line 5: RIP old friend

natemoo-re (approved):

kelvinz:
Is this fixed? I'm still having AlpineJS linting errors. =(

Princesseuh:
> Is this fixed? I'm still having AlpineJS linting errors. =(

Hello, this is currently available under the 0.29.0 Pre-Release. You can access it by pressing the "Switch to Pre-Release" button on the extension's page inside VS Code. For other editors, just updating the language-server package should be enough 

Since this is a major change, we're keeping it in the oven just a tad longer before releasing it publicly!

kelvinz:
I switched it to pre-release and restarted VS code, but it doesn't seem to be working still. T_T
`:id="$id('dropdown-button')"` still gives errors.

Princesseuh:
> I switched it to pre-release and restarted VS code, but it doesn't seem to be working still. T_T
> `:id="$id('dropdown-button')"` still gives errors.

Hmm, I think we might still have problems with attributes starting with `:`, I'll take a look, thank you!

Issue #296, opened 2022-05-22
🐛 BUG: In non-Astro workspaces, the extension warn about not being able to find Astro abusively 


Princesseuh:
### Describe the Bug

Currently, when opening Astro files outside of an Astro workspace, the extension will complain about not being able to find Astro. This is valid, however it's very annoying for users. It'd be great if the following happened:

- If an Astro file is opened inside a non-Astro workspace, we should warn once, and only once about not being able to find Astro and that the experience might be degraded
- When multiple workspaces (folders) are opened, we should check if the folder is an Astro project before warning. Otherwise, users see the warning for every single folder, which is very annoying, especially considering those folders are not even Astro related

Those two behavior are pretty undesirable, we should make sure the extension is helpful in cases where it matters and otherwise be as invisible as possible 

### Steps to Reproduce

1. Open an Astro file in an untitled workspace
2. Extension warn multiple times about not being able to find Astro

1. Open multiple folders, with one of them being an Astro project
2. Extension warn that Astro couldn't be found in every opened folder

Izzur:
Alternatively it would be helpful to add extension setting to disable Astro extension warning altogether

PR #343, opened 2022-07-19
Fix abusive warnings inside non-Astro workspaces

Princesseuh:
## Changes

Before checking if Astro is present in one of the opened workspaces, we now first check that it's actually an Astro workspace to see if our warning will be relevant or not

Fix https://github.com/withastro/language-tools/issues/296

## Testing

Tested manually + added a test for the utility

## Docs

N/A

### DIFF
--- a/packages/language-server/src/plugins/typescript/language-service.ts
+++ b/packages/language-server/src/plugins/typescript/language-service.ts
@@ -123,7 +123,10 @@ async function createLanguageService(
 
 	// Before Astro 1.0, JSX definitions were inside of the language-server instead of inside Astro
 	// TODO: Remove this and astro-jsx.d.ts in types when we consider having support for Astro < 1.0 unnecessary
-	if (astroVersion.major === 0 || astroVersion.full === '1.0.0-beta.0') {
+	if (
+		(astroVersion.major === 0 || astroVersion.full === '1.0.0-beta.0') &&
+		!astroVersion.full.startsWith('0.0.0-rc-') // 1.0.0's RC is considered to be 0.0.0, so we have to check for it
+	) {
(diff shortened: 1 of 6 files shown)


packages/language-server/src/server.ts
```diff
@@ -51,13 +57,6 @@ export function startLanguageServer(connection: vscode.Connection) {
 					type: MessageType.Warning,
 				});
 			}
-
-			if (astroVersion.exist && astroVersion.major === 0 && astroVersion.minor < 24 && astroVersion.patch < 5) {
```
Princesseuh, line 55: This is a leftover for <Astro 0.24.5 which was released around 5-6 months ago. I think it's fine to remove it now as we head into 1.0

packages/language-server/src/plugins/typescript/language-service.ts
```diff
@@ -123,7 +123,10 @@ async function createLanguageService(
 
 	// Before Astro 1.0, JSX definitions were inside of the language-server instead of inside Astro
 	// TODO: Remove this and astro-jsx.d.ts in types when we consider having support for Astro < 1.0 unnecessary
-	if (astroVersion.major === 0 || astroVersion.full === '1.0.0-beta.0') {
+	if (
+		(astroVersion.major === 0 || astroVersion.full === '1.0.0-beta.0') &&
+		!astroVersion.full.startsWith('0.0.0-rc-') // 1.0.0's RC is considered to be 0.0.0, so we have to check for it
```
Princesseuh, line 128: Unrelated to the PR really, just figured I'd fix that warning at the same time

This fix won't scale into the future of Astro 2.0, but, by then this code should hopefully have been removed 😓

matthewp (approved):

PR #348, opened 2022-07-25
Fix completions not working inside script tags, fix abusive completions

Princesseuh:
## Changes

This PR does two things:
- Fix the completion filtering logic so that we don't accidentally filter completions inside script tags
- Update the completion filtering logic so that some duplicated completions are now fixed and some uninteresting completions are now filtered

Additionally, this PR remove some now unnecessary logic that manually added extensions to auto imports inside script tags.

## Testing

Added tests for filtering completions in the start of components

## Docs

N/A

### DIFF
--- a/.changeset/brown-taxis-look.md
+++ b/.changeset/brown-taxis-look.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/language-server': patch
+---
+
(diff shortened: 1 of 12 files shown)


matthewp (approved):
lgtm!

PR #366, opened 2022-08-12
Solidify loading types from Astro

Princesseuh:
## Changes

~~- Bring Solid as a dependency~~
- Make our error handling when not being able to load types and modules from Astro and integrations much more robust. Previously, if the `@astrojs/svelte|vue` packages were installed but were too old / somehow broken, the language server would completely freeze up, now it'll show an error and keep going with its language server life
- Improve support for monorepos, weird workspaces, our monorepo. *Everything* should work in our monorepo now
- Much better support for workspaces where Astro is not installed, now we'll gracefully fallback to internal types that are of course not as good as the ones in Astro, but serviceable when just trying out `.astro` files or showcasing on streams etc. A notification will still be shown to warn user that the experience is degraded

## Testing

This is mostly covered by our tests already and they all pass!

## Docs

N/A

### DIFF
--- a/packages/language-server/types/astro-jsx.d.ts
+++ b/packages/language-server/types/astro-jsx.d.ts
@@ -1,16 +1,15 @@
 /// <reference lib="dom" />
 /* eslint @typescript-eslint/no-unused-vars: off */
 /**
- * Adapted from jsx-dom
- * @see https://github.com/proteriax/jsx-dom/blob/be06937ba16908d87bf8aa4372a3583133e02b8a/index.d.ts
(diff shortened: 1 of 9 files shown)


packages/language-server/types/astro-jsx.d.ts
```diff
@@ -1,16 +1,15 @@
 /// <reference lib="dom" />
```
Princesseuh, line 1: This file is copy-pasted from Astro's

natemoo-re (approved):
Looks great!

FredKSchott:
Types working in our monorepo... that's huge!

PR #375, opened 2022-08-15
Use Prettier and our plugin to format files

Princesseuh:
## Changes

Move from using VS Code's formatters to Prettier and our plugin, which going forward will be our official way to format Astro files. We bundle Prettier and the plugin in the language server unless the user has those installed in which case we load theirs

This should solve all the issues related to formatting people had, further issues related to formatting will be contained in https://github.com/withastro/prettier-plugin-astro/issues

## Testing

Modified tests for this. This is slightly inconvenient to test for in a monorepo like this, so we'll mostly test the basic things. Testing behaviour is left to the plugin itself

## Docs

Updated commands and README


.changeset/tidy-adults-arrive.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/language-server': minor
```
matthewp, line 2: should it be major?

Princesseuh: Language server is pre 1.0 still!

packages/language-server/src/importPackage.ts
```diff
@@ -48,3 +49,14 @@ export function importSvelteIntegration(fromPath: string): typeof svelte | undef
...
 }
+
+export function importPrettier(fromPath: string): typeof prettier {
+	// This shouldn't ever fail, because we bundle Prettier in the extension itself
+	const prettierPkg = getPackagePath('prettier', [fromPath, __dirname])!;
+
+	return require(prettierPkg);
+}
+
+export function importPrettierPlugin(fromPath: string): string {
```
matthewp, line 60: More of a function naming thing, but I find it a little unexpected that importPrettier imports it into the module graph but this one does not.

.changeset/tidy-adults-arrive.md
matthewp: Oh, right, forgot about that.

packages/language-server/src/importPackage.ts
matthewp: This isn't blocking to me, just an observation.

matthewp (approved):

PR #377, opened 2022-08-16
Automatically type `Astro.props` using the Props interface when available

Princesseuh:
## Changes

When a `Props` interface/type is available, we'll now type `Astro.props` on the Astro global with it automatically. The type is not strict, it's `Record<string, any> & Props`. We could make it strict, but it would require to copy the JSDoc definition of `Astro.props` in the language-server (and in the future, the compiler), which is fine but a bit annoying

## Testing

Added a test

## Docs

N/A

### DIFF
--- a/packages/language-server/src/plugins/typescript/astro2tsx.ts
+++ b/packages/language-server/src/plugins/typescript/astro2tsx.ts
@@ -3,12 +3,28 @@ import { parseAstro } from '../../core/documents/parseAstro';
 
 function addProps(content: string, className: string): string {
 	let defaultExportType = 'Record<string, any>';
+	let shouldAddGlobal = false;
+	let astroGlobal = "type AstroGlobal = import('astro').AstroGlobal";
+	const astroGlobalConstDef = `
+	/**
+	 * Astro global available in all contexts in .astro files
+	 *
+	 * [Astro documentation](https://docs.astro.build/reference/api-reference/#astro-global)
+	 */
+	declare const Astro: Readonly<AstroGlobal>;
+	`;
 
 	if (/(interface|type) Props/.test(content)) {
 		defaultExportType = 'Props';
+		shouldAddGlobal = true;
+		astroGlobal += ' & { props: Props }';
 	}
 
-	return EOL + `export default function ${className}__AstroComponent_(_props: ${defaultExportType}): any {}`;
+	return (
+		EOL +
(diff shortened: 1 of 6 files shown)


packages/language-server/test/plugins/typescript/features/SemanticTokensProvider.test.ts
```diff
@@ -73,48 +73,6 @@ describe('TypeScript Plugin#SemanticTokenProvider', () => {
 				type: TokenType.method,
 				modifiers: [TokenModifier.defaultLibrary],
 			},
-			{
```
Princesseuh, line 76: The Semantic Token test tested the Astro global with Astro.props, since this change the token type of that, let's just remove that part from the test

Since this test test other things and there's no reasons semantic tokens would fail on the Astro global in particular, this doesn't really remove coverage

packages/language-server/src/plugins/typescript/astro2tsx.ts
```diff
@@ -3,12 +3,28 @@ import { parseAstro } from '../../core/documents/parseAstro';
 
 function addProps(content: string, className: string): string {
 	let defaultExportType = 'Record<string, any>';
+	let shouldAddGlobal = false;
+	let astroGlobal = "type AstroGlobal = import('astro').AstroGlobal";
```
Princesseuh, line 7: This will work even in cases where Astro (the package, not Atlanta) is unavailable because `import('astro')` is then `any`

natemoo-re (approved):
👀 Love it!

zadeviggers:
To which interfaces does this apply? Would imported ones work? The docs say you need to export the interface, but that doesn't seem to be the case here.

Princesseuh:
> To which interfaces does this apply? Would imported ones work? The docs say you need to export the interface, but that doesn't seem to be the case here.

This type using the local definition named `Props`, so you can do the following with no issues:

```astro
---
import type { MySuperType } from "../something"

type Props = MySuperType

Astro.props // <--- This has type `Record<string, any> & MySuperType`
---
```

jasikpark:
Is the current behavior that unexported Props types would get applied to Astro.props?

Princesseuh:
> Is the current behavior that unexported Props types would get applied to Astro.props?

Yes, exporting the Props interface has never changed anything really. It's a leftover from docs from the beginning of Astro

jasikpark:
oh fair 😅 showing how long it's been since I've delved into that

PR #381, opened 2022-08-22
Fix regression related to completions for HTML attributes

Princesseuh:
## Changes

Fixed completions for HTML attributes not working anymore since 0.20.3. Attributes have no description, so we were filtering them out accidentally We didn't have a test for this, so we didn't catch that regression, unfortunately.

## Testing

Added a test

## Docs

N/A


tony-sull (approved):
lgtm!

PR #387, opened 2022-08-27
Updated TypeScript from 4.6.0 to 4.8.2

jonathantneal:
## Changes

- This updates the TypeScript dependencies from 4.6.0 to 4.8.2
- Starting in TypeScript 4.7.x, [`package.json` exports are supported](https://devblogs.microsoft.com/typescript/announcing-typescript-4-7/#package-json-exports-imports-and-self-referencing).
- During `build`, I noticed a missing interface for `InlayHintsOptions`, which was then added with citation from the TypeScript Fourslash suite.

## Testing

Followed CONTRIBUTING.md.

## Docs

dependency update only

### DIFF
--- a/.changeset/thick-cobras-pull.md
+++ b/.changeset/thick-cobras-pull.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/language-server': minor
+---
+
+Updated TypeScript from 4.6.0 to 4.8.2
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "eslint-plugin-prettier": "^4.2.1",
     "prettier": "^2.7.1",
     "turbo": "^1.0.0",
-    "typescript": "^4.6.0"
+    "typescript": "^4.8.2"
   },
   "engines": {
     "node": "^14.16.0 || >=16.0.0"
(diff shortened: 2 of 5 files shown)


Princesseuh:
Hello! Thank you for this PR. Unfortunately, updating to the latest TypeScript version is not that easy. 

Some changes to our module resolution cache and our code actions are needed to properly support 4.7. Additionally, our current language services force `moduleResolution` to `node` which is not ideal

jonathantneal:
Thanks for the quick review, @Princesseuh. To clarify, would the guidance be to apply similar kinds of changes as https://github.com/sveltejs/language-tools/pull/1526? I want to be sure I’m looking at the right change Svelte made. If you’re not sure, I can ask later this week, too. No rush, and again, thank you for the incredibly fast review.

Princesseuh:
> Thanks for the quick review, @Princesseuh. To clarify, would the guidance be to apply similar kinds of changes as https://github.com/sveltejs/language-tools/pull/1526? I want to be sure I’m looking at the right change Svelte made. If you’re not sure, I can ask later this week, too. No rush, and again, thank you for the incredibly fast review.

Yep! That's the right PR. Additionally, support for group aware organize imports should be considered for full support  https://github.com/sveltejs/language-tools/pull/1515

I'm okay with that one being tackled separately if you prefer not to, it's alright. I can tackle it myself låter

Princesseuh:
https://github.com/withastro/language-tools/pull/389 makes this PR unnecessary. We'll still need to add support for `node16` resolution and group aware imports, but it'll be tackled separately. Thank you for making this PR!

PR #389, opened 2022-08-29
Decouple TypeScript from the language server

Princesseuh:
## Changes

Big PR. This removes TypeScript as a direct dependency of the language server and instead load it dynamically from a initializing parameter. This unlock a lot of things:

- It makes the language server much much smaller, like 60mb smaller
- Allow us to bundle the language server which will make the VS Code extension faster, smaller and allow us to run it on the web
- Allows the user to control the TypeScript version / use VS Code version's like you would in .ts files
- Allow support for localized versions of TypeScript (so diagnostics are in the user's language)

image

## Testing

Tests all pass, tested VS Code parts manually (commands, locale etc), added a test for SignatureHelper

## Docs

Will inform other editors users independently from this PR that a new initialization parameter is needed now to get full support. The language server will fallback to HTML & CSS support only whenever it can't find TypeScript and show a notification, so hopefully migration isn't too bad


matthewp:
What was the decision on what to do about `astro check`?

Princesseuh:
> What was the decision on what to do about `astro check`?

We'll add TypeScript as a dep of Astro itself. From this PR, the `AstroCheck` constructor takes a parameter to specify the TypeScript instance to use, so we'll be able to refer to the one included in Astro

matthewp (approved):

Issue #390, opened 2022-08-29
🐛 BUG: Prettier settings from Prettier's VS Code extension are not loaded


Princesseuh:
### Describe the Bug

Some people like to configure their Prettier settings globally using the settings Prettier's VS Code extension offer. We should load those!

### Steps to Reproduce

1. Add some Prettier settings through the Prettier VS Code extension
2. Try to format
3. Settings are not respected!

PR #391, opened 2022-08-31
Load settings from the Prettier VS Code extension when it's available

Princesseuh:
## Changes

When available, we'll now load settings from the Prettier VS Code extension, (ex: `prettier.semi`). Our settings cascade matches the one of the Prettier VS Code extension in that if a Prettier config file is available, we'll completely ignore everything else. This should lead to a consistent experience between all files no matter if they use the Astro extension to format or the Prettier VS Code extension

Fix https://github.com/withastro/language-tools/issues/390

## Testing

Tested manually in addition to current tests

## Docs

N/A

### DIFF
--- a/.changeset/neat-turtles-sleep.md
+++ b/.changeset/neat-turtles-sleep.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/language-server': minor
+---
+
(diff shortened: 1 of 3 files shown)


natemoo-re (approved):
🙌🏻

Issue #212, opened 2022-03-16
🐛 BUG: OpenVSX deploy get stalled


Princesseuh:
### Describe the Bug

It has happened now a few times that our OpenVSX deploy get stalled permanently with no messages. 

If I were to guess, it might be because the extension is too large due to us not bundling it yet but this is just a guess, there's really not much info to go on :woman_shrugging: 

### Steps to Reproduce

1. Try to deploy a new version
2. Publish to OpenVSX task get stalled forever

Princesseuh:
Seems to be an upstream issue https://github.com/eclipse/openvsx/issues/471

PR #394, opened 2022-08-31
Bundle the language server inside the VS Code extension

Princesseuh:
## Changes

[VS Code heavily recommend bundling your extension](https://code.visualstudio.com/api/working-with-extensions/bundling-extension) for notably the following reasons: better performance, and compatibility with running your extension on the web and inside notebook renderers. So this does that.

This unlocks:
- Our extension now work inside the browser! (Partially, see caveat below)
- Our extension is now quicker to load, and also smaller! (1000+ files smaller, half the size)

**Known caveat:**
- Currently, certain features are disabled on the web because I couldn't figure out how to make them work. Those features are TypeScript and formatting. I'll investigate those more outside of this PR

First steps toward https://github.com/withastro/language-tools/issues/180
Might help with https://github.com/withastro/language-tools/issues/212

## Testing

There's no real change to the language-server itself, so this was tested manually through using the extension, both on node and on the web

## Docs

N/A


.github/workflows/release.yml
```diff
@@ -49,7 +49,10 @@ jobs:
         working-directory: ./packages/vscode
         run: |
           npm i
-          npx ovsx publish -p ${{ secrets.OVSX_TOKEN }}
+          npm run build:ci:node
+          npx ovsx publish -p ${{ secrets.OVSX_TOKEN }} --target win32-x64 win32-ia32 win32-arm64 linux-x64 linux-arm64 linux-armhf darwin-x64 darwin-arm64 alpine-x64 alpine-arm64
```
Princesseuh, line 53: OVSX's documentation is a bit confusing on this and I'm not sure if it's right. I'd expect so since it's supposed to be equivalent to VSCE, though

packages/ts-plugin/src/astro2tsx.ts
```diff
@@ -1,53 +1,49 @@
 import type { FileMapping } from './source-mapper';
-import { Project } from 'ts-morph';
```
Princesseuh, line 2: ts-morph is a super huge dependency, it didn't matter much before, but with all the changes in this PR, it's now twice as big as the entire extension + language-server and we don't use it for much so might as well remove it

packages/language-server/src/plugins/typescript/astro2tsx.ts
```diff
@@ -21,8 +20,8 @@ function addProps(content: string, className: string): string {
 	}
 
 	return (
-		EOL +
-		(shouldAddGlobal ? astroGlobal + EOL + astroGlobalConstDef : '') +
+		'\n' +
```
Princesseuh, line 23: As far as I could tell, it's not a problem to not use platform-specific line returns here. If it does, we it should be fairly easily to make a stub for it

packages/vscode/scripts/build-browser.js
```diff
@@ -0,0 +1,72 @@
...
+					return { path: pathBrowserify };
+				});
+				build.onResolve({ filter: /^util$/ }, (args) => ({
+					path: args.path,
+					namespace: 'util-ns',
+				}));
+				build.onLoad({ filter: /.*/, namespace: 'util-ns' }, () => ({
+					contents: `export const TextDecoder = {}`,
+					loader: 'js',
+				}));
```
Princesseuh, line 68: Despite only needing the type, VSCode's Emmet helper import from `util` without `import type`, which makes it pull a very very big polyfill (like, 2000 files large) of `util`. We don't need it anywhere else so we can just fake it to make it happy

matthewp (approved):

PR #395, opened 2022-08-31
Dont use workspace config when not supported

Princesseuh:
## Changes

We were wrongfully still using `workspace.getConfiguration` even on clients that didn't announce supporting `workspace/configuration`. This fix that

## Testing

Tested manually, we don't have a test setup for this right now (and all common clients support `workspace/configuration`

## Docs

N/A


natemoo-re (approved):
Nice!

PR #400, opened 2022-09-07
Add support for Node16 moduleResolution

Princesseuh:
## Changes

This adds support for TypeScript's 4.7 `moduleResolution: 'node16'`. Due to it being fairly more strict regarding extensions and whatnot, this requires changes (that really, are for the better!) in our module resolution logic

## Testing

Added a test

## Docs

N/A


packages/language-server/test/plugins/typescript/features/DocumentSymbolsProvider.test.ts
```diff
@@ -104,7 +104,7 @@ describe('TypeScript Plugin#DocumentSymbolsProvider', () => {
 				containerName: 'Template',
 				kind: SymbolKind.Constant,
 				location: {
-					range: Range.create(4, 9, 5, 25),
+					range: Range.create(4, process.platform === 'win32' ? 10 : 9, 5, 25), // NOTE: For some reason, the location of this is different on Windows, might be a bug?
```
Princesseuh, line 107: I don't know why that is, why it worked before and why only on that particular variable. Lots of mystery that I only vaguely feel like solving right now. It's only off by a single character, so functionality is still the same

packages/language-server/src/plugins/typescript/snapshots/SnapshotManager.ts
```diff
@@ -170,14 +170,12 @@ export class SnapshotManager {
...
-	getFileNames() {
-		return Array.from(this.documents.keys()).map((fileName) => toVirtualFilePath(fileName));
+	getFileNames(): string[] {
+		return Array.from(this.documents.keys());
 	}
 
-	getProjectFileNames() {
-		return this.projectFiles.map((file) => {
-			return toVirtualFilePath(file);
-		});
```
Princesseuh, line 180: So happy to remove this, it was the first nonsense I added to this codebase and now I'm finally removing it ☺️

matthewp (approved):

PR #405, opened 2022-09-08
Load web module from ?target=web

Princesseuh:
## Changes

Load the server web module using the `?target=web` query param to make sure it loads the proper version when distributed on the web (otherwise it loads the browser module of the node version, which is empty). In local with the web extension host the query param is just ignored, so there's no need to apply this conditionally 

## Testing

Tested manually

## Docs

Added some deprecation messages to the old formatting settings

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -32,7 +32,7 @@
     },
     {
       "name": "Launch Web Extension",
-      "type": "pwa-extensionHost",
+      "type": "extensionHost",
       "debugWebWorkerHost": true,
       "request": "launch",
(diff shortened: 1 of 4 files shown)


.vscode/launch.json
```diff
@@ -32,7 +32,7 @@
     },
     {
       "name": "Launch Web Extension",
-      "type": "pwa-extensionHost",
+      "type": "extensionHost",
```
Princesseuh, line 35: Apparently `pwa-extensionHost` is deprecated even though the documentation still refers to it? Not sure what's up there

Issue #407, opened 2022-09-12
💡 RFC: Add support for TypeScript 4.7 Group Aware imports


Princesseuh:
### Background & Motivation

Currently, our Organize Imports features is not compatible with TS 4.7's Group Aware imports. This would be really nice to implement!

### Possible solutions

I am admittedly not 100% sure what's needed for this at the moment. [Svelte's implementation could be looked at for inspiration](https://github.com/sveltejs/language-tools/commit/e3951f82599f370e3bbba56116670013affd13d1)

### Alternatives considered

N/A

### Risks, downsides, and/or tradeoffs

N/A

### Open Questions

N/A

### Help make it happen!

- [X] I am willing to submit a PR to implement this change.
- [ ] I am willing to submit a PR to implement this change, but would need some guidance.
- [ ] I am not willing to submit a PR to implement this change.

PR #408, opened 2022-09-13
Add test for group aware organize imports

Princesseuh:
## Changes

N/A. Just a test. Feature works

Fix https://github.com/withastro/language-tools/issues/407

## Testing

Added a test for TS 4.7 group aware organize imports

## Docs

N/A


yarn.lock
```diff
@@ -14,6 +14,24 @@
...
+    source-map "^0.7.3"
+    typescript "~4.6.4"
+    vscode-css-languageservice "^6.0.1"
+    vscode-html-languageservice "^5.0.0"
+    vscode-languageserver "^8.0.1"
+    vscode-languageserver-protocol "^3.17.1"
+    vscode-languageserver-textdocument "^1.0.4"
+    vscode-languageserver-types "^3.17.1"
+    vscode-uri "^3.0.3"
+
```
natemoo-re, line 34: Heads up! Looks like some package is not using the local version of `@astrojs/language-server`

yarn.lock
```diff
@@ -5649,6 +5667,11 @@ typescript@*:
   resolved "https://registry.npmjs.org/typescript/-/typescript-4.6.2.tgz"
   integrity sha512-HM/hFigTBHZhLXshn9sN37H085+hQGeJHJ/X7LpBWLID/fbc2acUMfU+lGD98X81sKP+pFa9f0DZmCwB9GnbAg==
 
+typescript@~4.6.4:
+  version "4.6.4"
+  resolved "https://registry.yarnpkg.com/typescript/-/typescript-4.6.4.tgz#caa78bbc3a59e6a5c510d35703f6a09877ce45e9"
+  integrity sha512-9ia/jWHIEbo49HfjrLGfKbZSuWo9iTMwXO+Ca3pRsSpbsMbc7/IU8NKdCZVRRBafVPGnoJeFL76ZOAA84I9fEg==
+
```
natemoo-re, line 5674: Also pulling in an old TypeScript version somehow

yarn.lock
Princesseuh: That package is `astro` itself, since it depends on a different version of the language-server (which in turn depends on a version of `typescript`)

PR #409, opened 2022-09-13
Migrate repo to pnpm

Princesseuh:
## Changes

Attempt to finally join the pnpm clan on the language-tools repo. VSCE and OSCE are still not compatible with it, so we still use npm there, unfortunately.

## Testing

We'll see if things work!

## Docs

N/A


.npmrc
```diff
@@ -0,0 +1,6 @@
+link-workspace-packages=true
+auto-install-peers=false
+
+# Those needs to be hoisted or the language-server tests will fail. Not sure why
+public-hoist-pattern[]=@vue/*
+public-hoist-pattern[]=svelte2tsx
```
Princesseuh, line 6: I don't really get why this is needed, I ended up finding this because `shamefully-hoist` would fix the issue and I narrowed it down to those packages.

Those packages are depended on in the respective's integrations editor endpoint, and the errors are weird when they're not hoisted, it's almost as if it was loading the wrong version?

packages/vscode/CONTRIBUTING.md
```diff
@@ -5,14 +5,8 @@
...
 
 ```shell
-yarn install
-yarn build
-```
-
-To start the development server run:
-
-```shell
-yarn dev:vscode
```
Princesseuh, line 15: This was outdated, you don't need to do this anymore

matthewp (approved):

PR #410, opened 2022-09-13
Fix astro check initializing before TypeScript was loaded

Princesseuh:
## Changes

We were running the checker init before loading TypeScript, oops

## Testing

Added a test for astro check

## Docs

N/A

### DIFF
--- a/packages/language-server/test/check.test.ts
+++ b/packages/language-server/test/check.test.ts
@@ -1,0 +1,36 @@
+import { AstroCheck } from '../src/check';
+import { DiagnosticSeverity, Range } from 'vscode-languageserver-types';
+import ts from 'typescript/lib/tsserverlibrary';
+import { pathToUrl } from '../src/utils';
+import { expect } from 'chai';
+import { join } from 'path';
+
+describe('astro check', async () => {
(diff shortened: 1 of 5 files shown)


packages/language-server/test/check.test.ts
```diff
@@ -0,0 +1,37 @@
...
+					{
+						code: 2304,
+						message: "Cannot find name 'doesntExist'.",
+						range: Range.create(1, 1, 1, 12),
+						severity: DiagnosticSeverity.Error,
+						source: 'ts',
+						tags: [],
+					},
+				],
+				filePath: new URL(pathToUrl(path)).pathname,
```
Princesseuh, line 32: This is necessary because Astro check currently handle paths the wrong way. It needs a minor to fix so I'll do that in the future

natemoo-re: This will break on Windows, IIRC. Are you looking for `pathToFileURL()`?

Princesseuh: This is actually exactly so the test pass on Windows haha. The language-server's part of `astro check` returns paths the wrong way, the Astro part of `astro check` fixes them back up for display.

PR #419, opened 2022-09-16
Improve completions on the Fragment element, add completions for slot on components

Princesseuh:
## Changes

Update completions on the Fragment element to not have client directives and the slot attribute. Also added the slot attributes to all components

Before:
image

After: 
image

## Testing

Updated tests

## Docs

N/A


matthewp (approved):
🙏

natemoo-re (approved):

PR #422, opened 2022-09-19
Add support for Go to References

Princesseuh:
## Changes

Add support for Go to References / Find References. This was a bit confusing because there's multiple ways to do this and I'm not sure the differences between them 🤔

Like for https://github.com/withastro/language-tools/pull/416, this is only the Astro part of it. The TS part will be tackled when we update the TS plugin.

## Testing

Added a test for this, added some tests that were missing from previous PRs

## Docs

N/A


matthewp (approved):
lgtm, love all the tests

PR #429, opened 2022-10-13
[WIP] Add basic support for CSS Module IntelliSense

Princesseuh:
## Changes

This PR adds basic intellisense for CSS modules. I'm not super satisfied with how the CSS classes are extracted, PostCSS isn't necessarily heavy but it's still a big process. Volar uses a regex to do this, but it doesn't seems to work perfectly either, so not sure if there's a perfect lightweight solution or not

image

## Testing

WIP

## Docs

N/A

### DIFF
--- a/packages/language-server/package.json
+++ b/packages/language-server/package.json
@@ -31,12 +31,16 @@
     "vscode-languageserver-protocol": "^3.17.1",
     "vscode-languageserver-textdocument": "^1.0.4",
     "vscode-languageserver-types": "^3.17.1",
-    "vscode-uri": "^3.0.3"
+    "vscode-uri": "^3.0.3",
+    "icss-utils": "^5.1.0",
+    "postcss": "^8.4.18",
+    "postcss-icss-selectors": "^2.0.3"
   },
   "devDependencies": {
     "@astrojs/svelte": "^1.0.0",
     "@astrojs/vue": "^1.0.0",
     "@types/chai": "^4.3.0",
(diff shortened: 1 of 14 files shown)


JulianCataldo:
Hey,

Here some randoms observations on CSS modules support in Astro extension.

With `clinyong.vscode-css-modules` we can get CSS peeking:

Screenshot 2022-10-15 at 17 35 36

In their demo  below, we can see **very** briefly an hint overlay with the class,
strange that we only have the CSS peek now (see above), no overlays anymore, this demo must be outdated:
l0EwY2Mk4IBgIholi

I suppose VS Code support hint / peeking works in a different way. 
If I had to choose between the two (like the CSS Modules shows), I think the CSS peeking is the most UX friendly, while it would be very neat to have both 😄

Anyway, regardless of Astro idiomatic CSS or CSS Modules choice, I think the click / peek to class is the most looked after feature, DX-wise, but I might be wrong.

Thx again for your work 👍.

Princesseuh:
Hey @JulianCataldo 

Not sure why it doesn't work in their case, but go to definition and peak definition uses the same information. 

You (normally) can't support only one of those. The implementation in this PR supports this (though the mapping are currently a bit wonky)

As for hover information, this PR currently does not have it, however it'd be possible to support fairly easily, I think

Princesseuh:
Closing since this will need to be implemented differently after https://github.com/withastro/language-tools/pull/502 (hopefully easier!)

PR #431, opened 2022-10-18
Add support for getting updated code from unsaved Astro, Svelte and Vue files

Princesseuh:
## Changes

Follow-up to https://github.com/withastro/language-tools/pull/423, now that I have a better understanding of how our snapshots are updated.

TypeScript in VS Code has this, in my opinion weird, feature where you don't need to actually save your files to get updated info about them. For instance, if you change the signature of a function in a file without saving and go to another file using it, the types will be already updated. 

As the title imply, we already supported this for TypeScript files, but the feature was unfortunately bugged for other support formats, this PR fix this

## Testing

Added tests and updated affected ones

## Docs

Unrelated to the main change, but this PR also add some documentation on our server's initialization parameters as we often get questions about them

### DIFF
--- a/.changeset/polite-frogs-deny.md
+++ b/.changeset/polite-frogs-deny.md
@@ -1,0 +1,6 @@
+---
+'@astrojs/language-server': patch
+'astro-vscode': patch
+---
+
+Add support for import completions of .md, .mdx and .html files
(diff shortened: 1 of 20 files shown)


natemoo-re (approved):
Interesting... cool feature!

PR #432, opened 2022-10-18
Add support for import path completions for .md, .mdx and .html files

Princesseuh:
## Changes

My TS plugin explorations in https://github.com/withastro/language-tools/pull/427 made me realize that it was very easy to do this, so here it is

image

I do believe `readDirectory` is used for other things, but I couldn't find any issues when testing both through our tests and manually

## Testing

Updated tests

## Docs

N/A


natemoo-re (approved):
Yessssss, this is great!

Issue #433, opened 2022-10-18
🐛 BUG: VSCode's `Organize Imports` command breaks `<script>` tags in `.astro` files


yegor-usoltsev:
### Describe the Bug

VSCode's `Organize Imports` command breaks `<script>` tags in `.astro` files.

For some reason `<script></script>` has been replaced with `<script>export { };` which leads to further compilation errors.

### Steps to Reproduce

1. Run `npm create astro@latest -- --template minimal`

2. Open project in Visual Studio Code, make sure [Astro Extension](https://marketplace.visualstudio.com/items?itemName=astro-build.astro-vscode) is installed and enabled

3. Add `<script></script>` tag to `src/pages/index.astro` file:

```diff
---
---

<html lang="en">
	<head>
		<meta charset="utf-8" />
		<link rel="icon" type="image/svg+xml" href="/favicon.svg" />
		<meta name="viewport" content="width=device-width" />
		<meta name="generator" content={Astro.generator} />
		<title>Astro</title>
+		<script></script>
	</head>
	<body>
		<h1>Astro</h1>
	</body>
</html>
```

4. Run `Organize Imports` command (<kbd>⇧</kbd><kbd>⌘</kbd><kbd>P</kbd> on macOS, type "Organize Imports", press <kbd>return</kbd>)

5. Expected no changes in the file, but got an invalid html:

```diff
---
---

<html lang="en">
	<head>
		<meta charset="utf-8" />
		<link rel="icon" type="image/svg+xml" href="/favicon.svg" />
		<meta name="viewport" content="width=device-width" />
		<meta name="generator" content={Astro.generator} />
		<title>Astro</title>
-		<script></script>
+		<script>export { };
+
	</head>
	<body>
		<h1>Astro</h1>
	</body>
</html>
```

```shell
$ npm run build

> @example/minimal@0.0.1 build
> astro build

12:32:47 AM [build] output target: static
12:32:47 AM [build] Collecting build info...
12:32:47 AM [build] Completed in 3ms.
12:32:47 AM [build] Building static entrypoints...
Unexpected character in scriptData: 
12:32:47 AM [build] Completed in 233ms.

 building client 
[vite:esbuild] Transform failed with 1 error:
/sandbox/my-astro-site/src/pages/index.astro?astro&type=script&index=0&lang.ts:3:2: ERROR: Unexpected "/"
file: /sandbox/my-astro-site/src/pages/index.astro?astro&type=script&index=0&lang.ts:3:2

[...]

PR #434, opened 2022-10-19
Fix Organize Imports adding virtual code to script tags

Princesseuh:
## Changes

In order to make them modules in TypeScript eyes, our script tags have some virtual code at the end. This does not cause issues in most cases, as we'll usually check to make sure positions are "in the real code" before doing anything. 

However, this doesn't always work for Organize Imports because it'll sometimes collapse code in a way where the beginning of the edit it gives is actually inside the real code, despite it mostly affecting virtual code. This goes both ways (edits to real imports will sometimes end in virtual code), so we can't check if both the start and the end of the edit is inside the real code without creating some situations where it wouldn't work

This PR works around this with our most reliable tool: `includes`.

Fix https://github.com/withastro/language-tools/issues/433

## Testing

Added tests for the specific issue and other possible edge cases

## Docs

N/A

### DIFF
--- a/.changeset/mighty-teachers-sneeze.md
+++ b/.changeset/mighty-teachers-sneeze.md
@@ -1,0 +1,6 @@
+---
+'@astrojs/language-server': patch
+'astro-vscode': patch
+---
+
+Fix Organize Imports sometimes adding code to script tags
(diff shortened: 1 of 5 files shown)


natemoo-re (approved):
Nice fix!

PR #435, opened 2022-10-19
Fix completions of component props not working if a CSS file was imported before the component

Princesseuh:
## Changes

Previously when looking for the component import for props, we'd give up too early if we found an import with no clause (ex: `import 'thing.css'`), so we wouldn't properly get the type of the component. This fix that

## Testing

Added test

## Docs

N/A


matthewp (approved):
nice test

bholmesdev (approved):
All these `if` statements have me feeling `clause`trophobic

PR #438, opened 2022-10-21
Fix numbers not being highlighted properly inside HTML attributes

Princesseuh:
## Changes

Numbers were not included in the regex, now they are!

## Testing

Added test

## Docs

N/A


natemoo-re (approved):
Nice catch!

PR #448, opened 2022-11-15
chore: upgrade compiler

natemoo-re:
## Changes

- Upgrades `@astrojs/compiler` to latest
- Fixes issues outlined in https://github.com/withastro/compiler/issues/631

## Testing

N/A, already covered

## Docs

N/A, dep bump


Princesseuh (changes requested):
Please update it in the TypeScript plugin package 😄

Princesseuh (approved):

PR #454, opened 2022-11-30
Fix prettier plugins not being loaded when formatting

Princesseuh:
## Changes

Previously we'd only load our own plugin when formatting, this PR make sure we also load other plugins. Notably, this fix compatibility with `prettier-plugin-tailwindcss`

## Testing

Tested manually

## Docs

N/A


aFuzzyBear (approved):
Hey sorry about the delay, this looks like a 🧠one, 
💚🌹

jasikpark (approved):
lgtm

PR #466, opened 2023-01-03
fix(completions): Fix completions not showing up in VS Code in certain cases

Princesseuh:
## Changes

VS Code filters completions on the client, both to avoid the server doing unnecessary work and also because it believes it can do it better than the servers (which is true in most cases). One of the way it filters them is by excluding any completions whose replacement value is somewhere unrelated to where the completion was asked. 

As expected, this filtering happens while building the completion list, which in the case of languages that needs to be mapped, the positions of the completions also needs to be mapped, or VS Code just won't show them. Mapping the list isn't super expensive and only happens in certain cases, so it should mostly be fine. The most of the completion is still spent inside TS itself, so, there's bigger fishes to fry for sure. If it turns out to be a problem, there's other ways to achieve a similar result that could be faster, I think

I believe this issue might be older than the new TSX output, but since the change to the output weren't as drastic back then, it was rarer.

This PR additionally remove some code that isn't needed with the new output that slipped through the initial PR for it

## Testing

This is unfortunately a case where we can't test it ourselves, because the issue inherently is in VS Code itself and not in our code. Our current tests covers our part, though.

## Docs

N/A

### DIFF
--- a/.changeset/chilly-points-sniff.md
+++ b/.changeset/chilly-points-sniff.md
@@ -1,0 +1,6 @@
+---
+'@astrojs/language-server': patch
+'astro-vscode': patch
+---
+
+Fix completions of strings not showing in certain cases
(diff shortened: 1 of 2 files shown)


natemoo-re (approved):
Not sure I fully understand why this is needed, but the code LGTM!

Issue #368, opened 2022-08-13
💡 RFC: Add support for renaming symbols 


Princesseuh:
I believe renaming has worked at some point, or at least in certain contexts at some point. Nowadays it just seems like it doesn't work at all, it'd be great to bring it back, both inside the HTML template and the JavaScript parts

Princesseuh:
This can be tackled right now, however the new TSX output will make it so this feature will need fairly consequential changes. As such, I suggest we wait for that to happen first before taking a serious look at renaming

PR #467, opened 2023-01-04
Add support for renaming

Princesseuh:
## Changes

Title is self explanatory! We had renaming support already, but it wasn't always quite working, especially on Windows

Fix https://github.com/withastro/language-tools/issues/368

## Testing

Added tests for it

## Docs

N/A


natemoo-re (approved):
Nice tests! Assuming everything passes, LGTM!

Issue #468, opened 2023-01-05
💡 RFC: Allow breakpoints in .astro files


Vap0r1ze:
### Background & Motivation

Using vitejs/vite#11576 and withastro/compiler#680, I'm able to use `debugger;` statements in frontmatter flawlessly.
![Code_4quUaJQw11](https://user-images.githubusercontent.com/20448879/210857945-cb15b59d-9d8e-4e35-8c31-046df50adc35.png)

The only missing piece of this puzzle now is the ability to create breakpoints in `.astro` files. I'm opening this issue here because I assume it's up to the LSP / extension to declare that the file can have breakpoints, and what lines/columns to allow them.

### Proposed Solution

Allowing breakpoints in `.astro` files

### Help make it happen!

- [ ] I am willing to submit a PR to implement this change.
- [X] I am willing to submit a PR to implement this change, but would need some guidance.
- [ ] I am not willing to submit a PR to implement this change.

Princesseuh:
Fair request! I think this is actually easy to add, just need to add the `breakpoints` field to the extension's package.json.

If you're interested in doing the PR, go ahead! Otherwise I'll take a look at it tomorrow 9AM (currently 22 in my timezone at the time of writing)

PR #470, opened 2023-01-06
feat(debugging): Add support for breakpoints inside Astro files

Princesseuh:
## Changes

This enables breakpoints inside Astro files, simple change. From my tests, it seems like they're not mapped correctly necessarily, but I don't think this is something we can control from the extension pov

Fix https://github.com/withastro/language-tools/issues/468

## Testing

N/A

## Docs

N/A

### DIFF
--- a/.changeset/old-humans-thank.md
+++ b/.changeset/old-humans-thank.md
@@ -1,0 +1,5 @@
+---
+'astro-vscode': patch
+---
+
+Add support for breakpoints
--- a/packages/vscode/package.json
+++ b/packages/vscode/package.json
@@ -129,6 +129,11 @@
         }
       ]
     },
+    "breakpoints": [
+      {
+        "language": "astro"
+      }
+    ],
     "configuration": {
       "type": "object",
       "title": "Astro configuration",


natemoo-re (approved):
LGTM! Would love to talk through the mapping issues.

Vap0r1ze:
> This enables breakpoints inside Astro files, simple change. From my tests, it seems like they're not mapped correctly necessarily, but I don't think this is something we can control from the extension pov

I'm not sure if my setup works completely, because after building and using the `Launch Extension` script, switching to an astro project gives me [weird syntax highlighting](https://user-images.githubusercontent.com/20448879/211072587-6c766a76-ee44-4cd9-86f3-981a87cd38bd.png)

But using the two pull requests I mentioned in my original issue, this PR seems to work pretty well:
<video src="https://user-images.githubusercontent.com/20448879/211081675-e0e4baa5-94ed-49fa-bf54-4747d2bd6713.mp4"/>

The only strange thing I noticed was the weird positioning of the breakpoints, seemingly almost overlapping with text. Also this double breakpoint here (I'm not sure if this is due to language-tools or compiler sourcemaps):
weird breakpoints

Princesseuh:
> > This enables breakpoints inside Astro files, simple change. From my tests, it seems like they're not mapped correctly necessarily, but I don't think this is something we can control from the extension pov
> 
> I'm not sure if my setup works completely, because after building and using the `Launch Extension` script, switching to an astro project gives me [weird syntax highlighting](https://user-images.githubusercontent.com/20448879/211072587-6c766a76-ee44-4cd9-86f3-981a87cd38bd.png)
> 
> But using the two pull requests I mentioned in my original issue, this PR seems to work pretty well:
> <video src="https://user-images.githubusercontent.com/20448879/211081675-e0e4baa5-94ed-49fa-bf54-4747d2bd6713.mp4"/>
> 
> The only strange thing I noticed was the weird positioning of the breakpoints, seemingly almost overlapping with text. Also this double breakpoint here (I'm not sure if this is due to language-tools or compiler sourcemaps):
> weird breakpoints
> 

You need to build the grammar files, I'll add it to the contributing guide!

PR #478, opened 2023-01-23
add when-clause for commands

### LINKED ISSUE #477: 🐛 bug(vscode): astro commands shows when not in astro files
### Describe the Bug

Most plugins, e.g. TypeScript only shows their "commands" if you're in a file or project which uses it - so the `TypeScript: Restart TypeScript server` only shows if I `CMD+SHIFT+P` in TypeScript files. This is nice since you only get relevant results when I search for the `restart` command.

CleanShot 2023-01-22 at 12 22 17@2x

However, this is not how Astro works. You get the `restart astro language server` no matter what file im browsing - so if I recently used the command it gets chosen even when im no longer in astro - if that makes sense... this is just "annoying" and "slows" down developers as we want to restart our TS servers quite "often".

CleanShot 2023-01-22 at 12 24 24@2x

> Note: I get that this is quite "nit" but it's still an annoying thing to have. The plugins commands doesn't need to show up when I'm not browsing Astro files.

### Steps to Reproduce

1. In any VSCode project, Restart the Astro language server by `CMD+shift+P` then type `restart` and select the astro one. (note that astro is the only one showing up (well you might have other plugins but at least the TS option doesn't show))
2. Later, when you're in a normal TS project, you want to restart the typescript server you do the same process - and muscle memory just does it without thinking, however, the Astro option will be chosen even though you're not in an Astro file =>> you have to do the process again and look what server you're restarting

juliusmarminge:
Closes #477

## Changes

- What does this change?

Adds a `when` clause to the commands so that they are only shown when browsing `.astro` files - which matches the behavior of the TypeScript plugin.

## Testing

TBD

## Docs

bug fix only - no docs needed

### DIFF
--- a/.changeset/kind-flowers-rush.md
+++ b/.changeset/kind-flowers-rush.md
@@ -1,0 +1,5 @@
+---
+'astro-vscode': patch
+---
+
+Fixed Astro commands showing even outside Astro files
--- a/packages/vscode/package.json
+++ b/packages/vscode/package.json
@@ -92,18 +92,22 @@
     "commands": [
       {
         "command": "astro.restartLanguageServer",
+        "when": "resourceExtName == .astro",
         "title": "Astro: Restart Language Server"
       },
       {
         "command": "astro.showTSXOutput",
+        "when": "resourceExtName == .astro",
         "title": "Astro: Debug: Show TSX Output"
       },
       {
         "command": "astro.selectTypescriptVersion",
+        "when": "resourceExtName == .astro",
         "title": "Astro: Select Typescript Version..."
       },
       {
         "command": "astro.findFileReferences",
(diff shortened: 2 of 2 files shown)


juliusmarminge:
Not very familiar with VSCode extensions so not sure how you test this? When I launch the extension from the debug menu, will the window opened use the local dev plugin then, and not the one I have installed globally?

Everything just died in that window xdddd, no syntax highlighting, no formatting etc so not sure I have the right setup..

CleanShot 2023-01-23 at 17 11 31@2x

Princesseuh (approved):
Thank you for the PR, I just moved the when clause to the proper area (you need to disable the command in the menu stuff, as the commands can exists but not show in the UI. It's weird)

PR #371, opened 2022-08-13
[feature] yarn berry support (#370)

ezracelli:
closes #370 

## Changes

- adds and implements `astro.langauge-server.ls-path` setting
- adds and implements `astro.language-server.runtime` setting

## Testing

Ran locally with the instructions from `./CONTRIBUTING.md` inside a project with PnP enabled

## Docs

No docs were updated, doesn't look like the settings are necessary documented anywhere inside the codebase

PR #480, opened 2023-02-06
Support local language servers in VS Code plugin

typedrat:
## Changes

Continues the work from #371 and yarnpkg/berry#5223 by enabling the Visual Studio Code plugin to load workspace-suggested language server versions after user confirmation to prevent security issues, as present in the Prettier and TypeScript extensions.

## Testing

Ran locally with a project that uses the Astro language server installed package-local with Yarn and it worked as expected.

## Docs

There is no existing documentation of this feature to update.


Princesseuh (approved):
Looks great, thank you!

PR #497, opened 2023-03-23
Add file templates

natemoo-re:
## Changes

- Add file templates for VS Code
- Helps users when starting a new file from a template

https://user-images.githubusercontent.com/7118177/227381303-e32f0f8e-8378-4b67-848b-4a6d06928fcc.mp4

## Testing

Tested manually

## Docs

Self-evident?


delucis:
Ooh cool! Re the self-evident docs bit, we do list all the supported features in https://github.com/withastro/language-tools/tree/main/packages/vscode#readme so maybe we should add this there?

packages/vscode/package.json
```diff
@@ -44,7 +44,6 @@
     "vscode": "^1.67.0"
   },
   "activationEvents": [
-    "onLanguage:astro",
```
Princesseuh, line 47: Any reason for this? Not all Astro workspaces have a Astro config file. (Ex: anonymous workspaces, fixtures etc)

packages/vscode/languages/astro.code-snippets
```diff
@@ -0,0 +1,40 @@
+{
+	"page_html": {
+		"prefix": "page_html",
+		"isFileTemplate": true,
+		"scope": "astro",
+		"body": [
+			"---",
+			"",
```
Princesseuh, line 8: Though on placing the first cursor here? I'm thinking the first cursor should be here and then the other ones should match Emmett's `doc` snippet

packages/vscode/package.json
natemoo-re: Ah. I was getting a lint warning that this was no longer needed in the latest versions, but probably wise to keep it for backwards compat.

Princesseuh: Ah yeah, I forgot about that change. They changed it in 1.74.0, but we support up to 1.67.0, so your assumption is correct that we should keep it!

Princesseuh (approved):

PR #482, opened 2023-02-10
1.0 for the VS Code extension, the TypeScript plugin and the Astro language server

Princesseuh:
## Changes

**Blocked by the following compiler issue:**
- https://github.com/withastro/compiler/issues/714

and the release of the 1.0 of the Prettier plugin:
- https://github.com/withastro/prettier-plugin-astro/pull/329

---

After a few months in pre-release squashing bugs left and right and making sure everything is stable, it's time to cut an official 1.0 release of all the packages in this repo! 

There's still a lot of work to be done in the language server, both on bugs and on features, but we generally consider the current experience to be stable and featureful enough to release a 1.0 and start looking toward the future.

After this, you can expect the packages in this repo to follow semver and adhere more officially to the process the core Astro follow for releasing new versions.

## Testing

N/A

## Docs

N/A

PR #504, opened 2023-04-11
chore: add repository and update prettier-plugin-astro

lvqq:
## Changes
- Add  repository in package.json
- Update prettier-plugin-astro, refer: https://github.com/withastro/astro/issues/6803

### DIFF
--- a/.changeset/ten-snakes-lie.md
+++ b/.changeset/ten-snakes-lie.md
@@ -1,0 +1,6 @@
+---
+'@astrojs/language-server': patch
+'astro-vscode': patch
+---
+
+update prettier-plugin-astro
--- a/packages/language-server/package.json
+++ b/packages/language-server/package.json
@@ -3,6 +3,11 @@
   "version": "0.29.6",
   "author": "withastro",
   "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/withastro/language-tools",
+    "directory": "packages/language-server"
+  },
   "type": "commonjs",
   "main": "dist/index.js",
   "bin": {
--- a/packages/language-server/package.json
+++ b/packages/language-server/package.json
@@ -30,7 +30,7 @@
     "@vscode/emmet-helper": "^2.8.4",
     "events": "^3.3.0",
     "prettier": "^2.7.1",
-    "prettier-plugin-astro": "^0.7.0",
+    "prettier-plugin-astro": "^0.8.0",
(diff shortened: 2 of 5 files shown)


Princesseuh:
It seems like there's now 4 versions of the compiler in the lockfile, it'd be great to untangle that while we're there, as otherwise it won't fix the situation happening in main

lvqq:
I agree. 

It could solve the issue in astro but will making things worse for this repository. What's the suggestion for this problem since there is a cycle in dependencies? 

@astro/language-server 0.29.6
  -> astro(devDependencies)
      -> @astro/language-server 0.28.3

And @astrojs/compiler is a direct dependency of @astro/language-server, is a indirect dependency of prettier-plugin-astro. For astro repository, a new version of @astro/language-server which depends on a v0.8.0 prettier-plugin-astro is needed

ematipico:
@lvqq 

You have few options here. You could start by breaking down this PR into two PRs: 
- one where you just update the `package.json` fields, which are only metadata;
- another when where you tackle the update of the dependency;

Regarding the update of the dependency:  

```
@astro/language-server 0.29.6
-> astro(devDependencies)
-> @astro/language-server 0.28.3(dependencies)
```

You could make `astro(devDependencies)` a dev and **optional** dependency. Then you could add the [`peerDependenciesMeta`](https://pnpm.io/package_json#peerdependenciesmetaoptional) to tell the package manager that the dependency is not mandatory. **I think** that should help.

Princesseuh:
I don't mind the devDeps at all actually. They're only used in tests to test that we get the types from Astro properly, so it doesn't really matter if we don't have a precise version for them. 

The main thing is avoiding multiple versions for users. Something to note is that `astro` cannot currently updates its `@astrojs/language-server` version until the later reaches 1.0 (https://github.com/withastro/language-tools/pull/482), because the versions of the package have diverged fairly consequentially.

Ideally, the path we'd take is the following:
- Release 1.0 of the Prettier plugin (https://github.com/withastro/prettier-plugin-astro/pull/329)
- Release 1.0 of the language-server (https://github.com/withastro/language-tools/pull/482), using the 1.0 of the Prettier plugin
- Both of them would use `^1.0.0` of `@astrojs/compiler`
- We update `astro` to use `^1.0.0` of the language-server

lvqq:
@Princesseuh Thanks for your patient reply! I'd move the modification about the repository to another PR

PR #509, opened 2023-04-17
feat: add ability to resolve `astro` in pnp workspace

Ty3uK:
## Changes

This change fix resolving `astro` package inside pnp workspace. With neovim and astro-lsp at this moment I have an error:

```
[START][2023-04-17 15:39:06] LSP logging initiated
[ERROR][2023-04-17 15:39:06] .../vim/lsp/rpc.lua:734	"rpc"	"astro-ls"	"stderr"	"/Users/ty3uk/projects/js/site seems to be an Astro project, but we couldn't find Astro or Astro is not installed\n"
```

This feature follows [Initializing PnP](https://yarnpkg.com/features/pnp#initializing-pnp) section and add `require` of `.pnp.cjs` file if it's exists in workspace.

## Testing

I've used custom `cmd` inside `nvim-lspconfig`:

```lua
lspconfig.astro.setup({
    on_attach = on_attach,
    capabilities = capabilities,
    cmd = {
        "node",
        "/Users/ty3uk/projects/js/astro-language-tools/packages/language-server/bin/nodeServer.js",
        "--stdio",
    },
```

## Docs

There is no need to update docs: it's working just from the box if user have `.pnp.cjs` file in workspace.


packages/language-server/src/server.ts
```diff
@@ -46,11 +53,22 @@ export function startLanguageServer(connection: vscode.Connection, env: RuntimeE
 	let hasConfigurationCapability = false;
 
 	connection.onInitialize((params: vscode.InitializeParams) => {
+		let isPnpInit = false;
```
Ty3uK, line 56: Not sure can we `require` this file multiple times for each workspace in context of a single server instance (will each import change previous behavior?)

packages/language-server/src/utils.ts
```diff
@@ -294,3 +294,14 @@ export function getAstroInstall(basePaths: string[]): AstroInstall | undefined {
 		},
 	};
 }
+
+export function getWorkspacePnpPath(workspacePath: string): string | null {
+	const possiblePath = resolve(workspacePath, '.pnp.cjs');
+
+	try {
+		require.resolve(possiblePath);
```
Ty3uK, line 302: First try was using `fs.existsSync`, but `esbuild` was complaining about `node` environment in browser build, so I decided to use that method.

packages/language-server/src/server.ts
```diff
@@ -46,11 +53,22 @@ export function startLanguageServer(connection: vscode.Connection, env: RuntimeE
...
+			let possiblePnpPath = null;
+
+			if (!isPnpInit) {
+				possiblePnpPath = getWorkspacePnpPath(uri);
+
+				if (possiblePnpPath) {
+					require(possiblePnpPath).setup();
+					isPnpInit = true;
+				}
+			}
```
ematipico, line 71:
```suggestion
			if (!isPnpInit) {
				const possiblePnpPath = getWorkspacePnpPath(uri);

				if (possiblePnpPath) {
					require(possiblePnpPath).setup();
					isPnpInit = true;
				}
			}
```

packages/language-server/src/utils.ts
```diff
@@ -294,3 +294,14 @@ export function getAstroInstall(basePaths: string[]): AstroInstall | undefined {
 		},
 	};
 }
+
+export function getWorkspacePnpPath(workspacePath: string): string | null {
+	const possiblePath = resolve(workspacePath, '.pnp.cjs');
```
ematipico, line 299: This operation can throw an error too, we should move it inside the `try/catch` block

Ty3uK: Done, thanks!

Ty3uK:
@ematipico, check PR, please 🙂

Princesseuh:
We'll be reviewing once we're back to work on Tuesday, thank you for your contribution!

Ty3uK:
> We'll be reviewing once we're back to work on Tuesday, thank you for your contribution!

No rush, thank you so much!

packages/language-server/src/server.ts
```diff
@@ -46,11 +53,20 @@ export function startLanguageServer(connection: vscode.Connection, env: RuntimeE
...
 		const environment: 'node' | 'browser' = params.initializationOptions?.environment ?? 'node';
 		const workspaceUris = params.workspaceFolders?.map((folder) => folder.uri.toString()) ?? [params.rootUri ?? ''];
 
 		workspaceUris.forEach((uri) => {
 			uri = urlToPath(uri) as string;
+			if (!isPnpInit) {
+				const possiblePnpPath = getWorkspacePnpPath(uri);
+
+				if (possiblePnpPath) {
+					require(possiblePnpPath).setup();
```
Princesseuh, line 66: This is a security issue, the server will execute any file named `.pnp.cjs` and its `setup` method. 

I am unfamiliar with PnP, but this should be either behind an init flag or only done after a question has been asked to the user.

Princesseuh (changes requested):
For security purpose, we should only do this with user approval. Additionally, this needs to be locked behind workspace trust when applicable

packages/language-server/src/server.ts
Ty3uK: That's an interesting question. My mind on the flag side, because we're talking about lsp server and I don't know how to actually show prompt to user from server's side

Princesseuh: I am okay with an init flag, it's *technically* possible to show a prompt and answer from the server's side (at least, VS Code can do it), but it's super cumbersome so

Ty3uK: Actually, I've figured out that this feature isn't so hard to implement 🙈
image
Thanks for feedback, I'll try this approach.

Ty3uK: @Princesseuh, so I've decided to use explicit init option.

Notification with actions returns a Promise, so in existing architecture it's hard to get the result without rewriting at least half of `server.ts` file 🙈

With this option i've got this in my lsp config:

image

Princesseuh:
Last thing would be to add the init params to https://github.com/withastro/language-tools/blob/main/docs/language-server/initParams.md, so people writing plugins can know that this setting exists

Ty3uK:
> Last thing would be to add the init params to https://github.com/withastro/language-tools/blob/main/docs/language-server/initParams.md, so people writing plugins can know that this setting exists

Done! Can you check correctness, please? English isn't my native language 🙈

Princesseuh (approved):
Looks great! Thank you for contributing this

Ty3uK:
> Looks great! Thank you for contributing this

Thanks for awesome feedback!

Issue #526, opened 2023-05-05
🐛 BUG: Language Server request contains path instead of URI for workspace/configuration


beeb:
### Describe the Bug

The [LSP specification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_configuration) dictates that the `workspace/configuration` method request contains a `scopeUri` which should be of type `DocumentUri`.

The astro language server apparently sends a path, when looking at the logs of my editor (helix):

```
2023-05-05T09:50:16.117 helix_term::application [DEBUG] received editor event: LanguageServerMessage((0, MethodCall(MethodCall { jsonrpc: Some(V2), method: "workspace/configuration", params: Map({"items": Array [Object {"scopeUri": String("/home/valentin/[redacted]"), "section": String("astro.typescript")}]}), id: Num(1) })))
2023-05-05T09:50:16.117 helix_term::application [ERROR] Language Server: Received malformed method call workspace/configuration in request 1: protocol error: InvalidParams: Invalid params: invalid value: string "/home/valentin/[redacted]", expected relative URL without a base.
2023-05-05T09:50:16.117 helix_lsp::transport [INFO] -> {"jsonrpc":"2.0","error":{"code":-32700,"message":"Malformed method call: workspace/configuration"},"id":1}
```

### Steps to Reproduce

N/A

### System

WSL2 (Ubuntu) on Windows 11

beeb:
Potentially the bug lies here:

https://github.com/withastro/language-tools/blob/750d69b03156c2ea200a8e8d1b60ebf684ca04a9/packages/language-server/src/plugins/typescript/language-service.ts#L114

beeb:
@Princesseuh I had some time and it was an easy fix: https://github.com/withastro/language-tools/pull/527

It indeed makes it work with helix editor after that change.

PR #527, opened 2023-05-05
Fix `workspace/configuration` request parameter type

beeb:
Closes https://github.com/withastro/language-tools/issues/526

## Changes

The [LSP specification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_configuration) dictates that the `workspace/configuration` method request contains a `scopeUri` which should be of type `DocumentUri`. This request was sometimes being sent with a path instead.

## Testing

Bugfix, not sure how to test

## Docs

No documentation update necessary

### DIFF
--- a/.changeset/modern-months-stare.md
+++ b/.changeset/modern-months-stare.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/language-server': patch
+---
+
+Fix type for `scopeUri` in workspace/configuration request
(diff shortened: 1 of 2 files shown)


ematipico:
> Bugfix, not sure how to test

Have you tried to install the extension locally and test it? That could help!

beeb:
@ematipico I tested the language server installed locally with helix editor and that worked! Not using VSCode at the moment.

Princesseuh:
This looks fine to me! However, the tests failure on Node 16 looks like they're valid ones. Could you look into it? Otherwise, I can look into it myself

beeb:
@Princesseuh OK the error is due to the `workspacePath` being null in the test case, which doesn't play nice with `pathToUrl`. No idea why the `workspacePath` is null but I added a check which now makes the test pass.

beeb:
Gotta love typescript's "type safety" 🙄

beeb:
If I had to take a guess, I would say that `workspaceUris` contains an entry that is not a `file://` URI, which makes `urlToPath` return `null` on line 109. But since there is `as string` the error is ignored.

Issue #519, opened 2023-05-01
import type from astro file to typescript file


nirus:
### What version of `astro` are you using?

astro@2.3.2

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

yarn

### What operating system are you using?

Mac

### What browser are you using?

Chrome

### Describe the Bug

I am unable to import the type declared in `.astro` file to `typescript` code in VScode editor. 

Below error is being displayed

<img width="1111" alt="image" src="https://user-images.githubusercontent.com/2213751/235432166-3227e1b9-577c-4721-9c0a-f588242baa54.png">

I upgraded the `package.json` using `yarn upgrade` to use latest packages. Still seeing the error shown in the screenshot.

I have attached the whole repo with the code and config settings.

### Link to Minimal Reproducible Example

https://github.com/nirus/cr-engine/blob/main/src/restapi/fetchAuthor.ts#L4

### Participation

- [ ] I am willing to submit a pull request for this issue.

nirus:
is this fixed?

Princesseuh:
> is this fixed?

Another unrelated fix for it will be released later today, which should hopefully fix all the issues.

PR #532, opened 2023-05-10
fix: Fix TypeScript plugin compatibility with TS 5.0

Princesseuh:
## Changes

Our plugin was broken in TS 5.0+ due to internal changes to TS. This PR fixes this by using the recommended method for patching language services inside TypeScript plugins and patching the right method for resolving modules

Fix https://github.com/withastro/language-tools/issues/519

## Testing

Tests should pass!

## Docs

N/A

### DIFF
--- a/packages/ts-plugin/src/language-service/index.ts
+++ b/packages/ts-plugin/src/language-service/index.ts
@@ -10,19 +10,47 @@ import { decorateGetImplementation } from './implementation.js';
 import { decorateLineColumnOffset } from './line-column-offset.js';
 import { decorateRename } from './rename.js';
 
+const astroPluginPatchSymbol = Symbol('astroPluginPatchSymbol');
+
+export function isPatched(ls: ts.LanguageService) {
+	return (ls as any)[astroPluginPatchSymbol] === true;
+}
+
 export function decorateLanguageService(
 	ls: ts.LanguageService,
 	snapshotManager: AstroSnapshotManager,
 	logger: Logger
 ): ts.LanguageService {
-	decorateLineColumnOffset(ls, snapshotManager);
-	decorateRename(ls, snapshotManager, logger);
-	decorateDiagnostics(ls);
-	decorateFindReferences(ls, snapshotManager, logger);
-	decorateCompletions(ls, logger);
-	decorateGetDefinition(ls, snapshotManager);
-	decorateGetImplementation(ls, snapshotManager);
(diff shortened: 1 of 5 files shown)


packages/ts-plugin/src/language-service/index.ts
```diff
@@ -10,19 +10,46 @@ import { decorateGetImplementation } from './implementation.js';
 import { decorateLineColumnOffset } from './line-column-offset.js';
 import { decorateRename } from './rename.js';
 
+const astroPluginPatchSymbol = Symbol('astroPluginPatchSymbol');
```
matthewp, line 13: Approved on the spot.

matthewp (approved):

packages/ts-plugin/src/language-service/index.ts
Princesseuh: Unfortunately not my code, copied that part from Svelte's TS plugin, sorry to disappoint

space-otter:
Unfortunately, the error persists. Export and import statments in `.ts` files don't recognize `.astro` files.

```
Cannot find module './Perk.astro' or its corresponding type declarations.
```

Princesseuh:
> Unfortunately, the error persists. Export and import statments in `.ts` files don't recognize `.astro` files.
> 
> ```
> Cannot find module './Perk.astro' or its corresponding type declarations.
> ```

Do you have a project I can look at?

space-otter:
> > Unfortunately, the error persists. Export and import statments in `.ts` files don't recognize `.astro` files.
> > ```
> > Cannot find module './Perk.astro' or its corresponding type declarations.
> > ```
> 
> Do you have a project I can look at?

For example an actual repo (https://github.com/prismaticweb/mmw-webpage/blob/main/src/components/PerkCollection/index.ts). I tried to install the ts-plugin and add it manually, without the plugin, switching pre-release, etc.

Princesseuh:
Thank you, I'll take a look!

Princesseuh:
Found the issue, will send out a fix as soon as possible. The plugin actually does work, I think there's just a packaging error

space-otter:
> Found the issue, will send out a fix as soon as possible. The plugin actually does work, I think there's just a packaging error

Uh, perfect! Thank you for the quick replay and fix.

Princesseuh:
This PR should fix the issue: https://github.com/withastro/language-tools/pull/538. Sorry for the inconvenience!

PR #538, opened 2023-05-15
fix: typescript imports in the ts-plugin

Princesseuh:
## Changes

After https://github.com/withastro/language-tools/pull/532, we had some random TypeScript imports in the code. We can't have that, we should always use the TypeScript that's passed down to us from the `src/index.ts`, because we don't necessarily have TypeScript in the user's project.

## Testing

Tested manually, can't really test this in our monorepo, we always have access to TypeScript, that's why it wasn't caught previously.

## Docs

N/A

### DIFF
--- a/.changeset/gentle-houses-know.md
+++ b/.changeset/gentle-houses-know.md
@@ -1,0 +1,6 @@
+---
+'@astrojs/ts-plugin': patch
+'astro-vscode': patch
+---
+
+Fix TypeScript plugin crashing at start in certain circumstances


MoustaphaDev (approved):
I'm not familiar with this repo but the code looks right to me!

packages/ts-plugin/src/language-service/diagnostics.ts
```diff
@@ -1,13 +1,15 @@
-import ts from 'typescript/lib/tsserverlibrary';
+import type ts from 'typescript/lib/tsserverlibrary';
 import { isAstroFilePath } from '../utils.js';
 
+type _ts = typeof ts;
```
jasikpark, line 4: why this rename? what does it signify?

jasikpark: ohh it's a type instead of a variable

jasikpark (commented):
lgtm?

matthewp (approved):

PR #561, opened 2023-06-05
fix: changed the LanguageClient.id because trace.server is not working

yaegassy:
## Changes

Regarding the configuration of `astro.trace.server`. 

https://github.com/withastro/language-tools/blob/eb2ce0b3effd00e49294be57ef0f3d1cb5f8f454/packages/vscode/package.json#L141

I have made modifications because there seems to be a difference between the configured LanguageClient.id and the actual behavior.

https://github.com/withastro/language-tools/blob/eb2ce0b3effd00e49294be57ef0f3d1cb5f8f454/packages/vscode/src/client.ts#L69

## Testing

None

## Docs

None

### DIFF
--- a/.changeset/happy-geese-invent.md
+++ b/.changeset/happy-geese-invent.md
@@ -1,0 +1,5 @@
+---
+'astro-vscode': patch
+---
+
+Fix `astro.trace.server` not working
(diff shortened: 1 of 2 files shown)


Princesseuh (approved):
Awesome, thank you!

PR #564, opened 2023-06-05
fix: pin the @volar/* and volar-service-* versions to avoid regression

intagaming:
According to semver, API changes shouldn't happen via a minor upgrade but it did, hence this change.

## Changes

- Pin some packages due to unexpected API changes.

## Testing

Tested locally via `npm pack` and loaded the package via the packed file.

## Docs

No documentation needed, just need some package version pins.


Princesseuh (approved):
Looks great to me, thank you! Hopefully we can upgrade to 1.7.0 once it's fully stable :)

packages/language-server/package.json
```diff
@@ -29,18 +29,18 @@
     "@astrojs/compiler": "^1.5.0",
     "@jridgewell/sourcemap-codec": "^1.4.15",
     "@volar/kit": "^1.6.9",
```
johnsoncodehk, line 31: You missed this. 👀

intagaming: Damn...

PR #592, opened 2023-06-28
Migrate the TS Plugin to Volar

johnsoncodehk:
## Changes

Migrated to TS Plugin integrated by Volar, this improved some renaming results execute from .ts.


johnsoncodehk:
Make @astrojs/compiler external in https://github.com/withastro/language-tools/pull/592/commits/a82ab803326ebba212e1d729e6fab31ff7a5ee07, so we don't need to copy `astro.wasm`, and the extension size is reduced to 5.1 MB.

ematipico:
Thank you @johnsoncodehk  CI is failing, it can't find a file. Could you look at that please?

johnsoncodehk:
Should be working now, please try again. :) Now the extension size reduce to 3.8 MB.

There are 2 breaking changes from ts plugin:

1. Unless .astro has been imported by a .ts file, tsserver does not know it, because the volar ts plugin only adds it to the LS host when it actually parses the .astro file path, not in advance from the project in getExternalFiles Search for all .astro files.
In addition to better performance, the ts plugin does not need to care about whether the current astro project or not.
This makes the auto import component unavailable for .astro files that were never imported. This is not a problem for Vue, because we only care about renaming, find reference and other functions.
If Astro still needs auto-import functionality, you has the option to implement getExternalFiles again.

2. Now `"compilerOptions": { "jsx": "preserve" }` needs to be set by the user, but Astro can also choose to add back the logic of when the current project is Astro, modify getCompilationSettings.

roninjin10:
Should the “work in progress” be removed from the readme?

matthewp:
> Should the “work in progress” be removed from the readme?

Yep! I think so.

Princesseuh:
> 1. Unless .astro has been imported by a .ts file, tsserver does not know it, because the volar ts plugin only adds it to the LS host when it actually parses the .astro file path, not in advance from the project in getExternalFiles Search for all .astro files.
>    In addition to better performance, the ts plugin does not need to care about whether the current astro project or not.
>    This makes the auto import component unavailable for .astro files that were never imported. This is not a problem for Vue, because we only care about renaming, find reference and other functions.
>    If Astro still needs auto-import functionality, you has the option to implement getExternalFiles again.

Hmm, it seems like it's not just the auto-import though, for renaming and file references to work, the `.astro` file needs to have been imported somewhere in a `.ts(x)` file. It's super annoying, but think I might have to implement `getExternalFiles` again for this

johnsoncodehk:
Regarding 1., since there are some projects like https://github.com/mizdra/typescript-plugin-asset that highly require auto-import, I'll investigate again whether this should/could be supported by the framework.

johnsoncodehk:
> it seems like it's not just the auto-import though, for renaming and file references to work

These problems have been solved, please receive the goods. 🤞

Issue #616, opened 2023-07-27
🐛 BUG: formatting .astro files throwing error in vscode


runeb:
### Describe the Bug

Invoking "Format Document" does nothing to the file and throws an error in the output for Astro Language Server

```
Error [ERR_UNSUPPORTED_DIR_IMPORT]: Directory import '/Users/runeb/src/monorepo/node_modules/prettier-plugin-astro' is not supported resolving ES modules imported from /Users/runeb/src/monorepo/node_modules/prettier/index.mjs
Did you mean to import /Users/runeb/src/monorepo/node_modules/prettier-plugin-astro/dist/index.js?
    at new NodeError (node:internal/errors:387:5)
    at finalizeResolution (node:internal/modules/esm/resolve:425:17)
    at moduleResolve (node:internal/modules/esm/resolve:1012:10)
    at defaultResolve (node:internal/modules/esm/resolve:1220:11)
    at nextResolve (node:internal/modules/esm/loader:165:28)
    at ESMLoader.resolve (node:internal/modules/esm/loader:844:30)
    at ESMLoader.getModuleJob (node:internal/modules/esm/loader:431:18)
    at ESMLoader.import (node:internal/modules/esm/loader:528:22)
    at importModuleDynamically (node:internal/modules/esm/translators:110:35)
    at importModuleDynamicallyCallback (node:internal/process/esm_loader:35:14) {
  code: 'ERR_UNSUPPORTED_DIR_IMPORT',
  url: 'file:///Users/runeb/src/monorepo/node_modules/prettier-plugin-astro'
}
```

I have attempted to re-install the Astro extension and clear out node_modules reinstalling everything

### Steps to Reproduce

1. Initialize an Astro project as an app in a [turbo repo](https://turbo.build/repo) at `apps/web`
3. Use vim mode in vscode (no idea if its relevant)
4. Install the latest (v2.1.3) Astro vscode plugin and attempt to format an .astro file
5. See that nothing changes in the file and no error message displayed
6. Open the Output tab of vscode and see the error pasted above

runeb:
Being careful to remove any local `prettier-plugin-astro` modules, the error is the same, but with different paths.

I also upgraded to `astro@2.9.3`

```
[...]

PR #617, opened 2023-07-28
fix(formatting): Fix Prettier not importing the right file when formatting 

Princesseuh:
## Changes

`getPackagePath` returns the root of the package (since that's what we want for editor integrations), however since Prettier plugins are now ESM, what we need is the specific file to import, so this fix that

Fix https://github.com/withastro/language-tools/issues/616

## Testing

I added a test 

## Docs

N/A

### DIFF
--- a/packages/language-server/src/languageServerPlugin.ts
+++ b/packages/language-server/src/languageServerPlugin.ts
@@ -75,8 +75,7 @@ export const plugin: LanguageServerPlugin = (
 					languages: ['astro'],
 					ignoreIdeOptions: true,
 					resolveConfigOptions: {
-						// Prettier's cache is a bit cumbersome, because you need to reload the config yourself on change
-						// TODO: Upstream a fix for this
+						// This seems to be broken since Prettier 3, and it'll always use its cumbersome cache. Hopefully it works one day.
 						useCache: false,
 					},
 					additionalOptions: async (resolvedConfig) => {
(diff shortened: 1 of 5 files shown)


packages/language-server/src/languageServerPlugin.ts
```diff
@@ -75,8 +75,7 @@ export const plugin: LanguageServerPlugin = (
 					languages: ['astro'],
 					ignoreIdeOptions: true,
 					resolveConfigOptions: {
-						// Prettier's cache is a bit cumbersome, because you need to reload the config yourself on change
-						// TODO: Upstream a fix for this
+						// This seems to be broken since Prettier 3, and it'll always use its cumbersome cache. Hopefully it works one day.
```
Princesseuh, line 78: Prettier 3 seems to have a weird caching issue. I tried everything: Setting `useCache` to `false`, calling `prettier.clearConfigCache()` manually etc and nothing worked, it always uses a cached config. This is only annoying if you're updating your Prettier config often, which hopefully doesn't happen too much.

packages/language-server/test/misc/format.test.ts
```diff
@@ -0,0 +1,27 @@
...
+describe('Formatting', () => {
+	let languageServer: LanguageServer;
+
+	before(async () => {
+		languageServer = await getLanguageServer();
+	});
+
+	it('Can format document', async () => {
+		const document = await languageServer.helpers.openFakeDocument(`---\n\n\n---`);
+		const formatEdits = await languageServer.helpers.requestFormatting(document, {
```
Princesseuh, line 15: Those APIs are so clean, so neat to use. Good job me!

matthewp (approved):

PR #628, opened 2023-08-21
Fix wordpattern syntax in astro-language-configuration.json

sebthom:
## Changes

According https://code.visualstudio.com/api/language-extensions/language-configuration-guide#word-pattern `wordPattern` must be a string, not an object.

This PR changes the `wordPattern` to string to comply with the spec.

## Testing

Tested manually.

## Docs

Bug Fix only


Princesseuh (approved):
Weird, I think I copied that snippet from VS Code's HTML configuration. Thank you for fixing it!

PR #632, opened 2023-08-23
feat: Add new completions for getStaticPaths, interface Props and prerender

Princesseuh:
## Changes

Some cute snippets completions for a neat DX.

https://github.com/withastro/language-tools/assets/3019731/87d97b15-ce68-4161-8130-7e5a66ecd42c

For the moment this only work in Astro files. We need to make some changes to our TS plugin before we can make that work there

## Testing

Added a test

## Docs

N/A


matthewp (approved):
fun!

ematipico (approved):

PR #634, opened 2023-08-23
Support prettier-plugin-astro v0.12 in language-server

bluwy:
## Changes

Bump peer dep range to support prettier-plugin-astro 0.12. Encountered this while bumping deps in core.

## Testing

Ran formatting and checks in core with 0.12 and it works.

## Docs

n/a. bump range only.

### DIFF
--- a/packages/language-server/package.json
+++ b/packages/language-server/package.json
@@ -61,7 +61,7 @@
   },
   "peerDependencies": {
     "prettier": "^3.0.0",
-    "prettier-plugin-astro": "^0.11.0 || ^0.12.0"
+    "prettier-plugin-astro": ">=0.11.0"
   },
   "peerDependenciesMeta": {
     "prettier": {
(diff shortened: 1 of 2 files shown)


packages/language-server/package.json
```diff
@@ -61,7 +61,7 @@
   },
   "peerDependencies": {
     "prettier": "^3.0.0",
-    "prettier-plugin-astro": "^0.11.0"
+    "prettier-plugin-astro": "^0.11.0 || ^0.12.0"
```
Princesseuh, line 64: Since it's unlikely that the API will change, I think we could even use `>=0.11.0`

bluwy: If you're confident about it, I'm also fine too 😄 
```suggestion
    "prettier-plugin-astro": ">=0.11.0"
```

Princesseuh (approved):

PR #653, opened 2023-09-19
Add a new troubleshooting item about formatting

Khaledgarbaya:
## Changes

- What does this change?
- Added a new troubleshooting entry about formatting `.astro` files because quite a few folks are having this [problem](https://x.com/PaulieScanlon/status/1701583756576858485?s=20)

## Testing

Yes Tested on vs code and other [folks](https://x.com/PaulieScanlon/status/1701583756576858485?s=20) tried it too
## Docs
No sure?

### DIFF
--- a/packages/vscode/README.md
+++ b/packages/vscode/README.md
@@ -26,8 +26,19 @@ A TypeScript plugin adding support for importing and exporting Astro components 
 
 HTML, CSS and TypeScript settings can be configured through the `html`, `css` and `typescript` namespaces respectively. For example, HTML documentation on hover can be disabled using `'html.hover.documentation': false`. Formatting can be configured through [Prettier's different configuration methods](https://prettier.io/docs/en/configuration.html).
 
+
+
 ## Troubleshooting
 
+### Auto-formatting for `.astro` files does not work
+
+To enable auto-formatting on VS code for `.astro` files make sure to add this line to your vs code settings:
+
+```
+"[astro]": {
+"editor.defaultFormatter": "astro-build.astro-vscode"
+}
+```
 ### Inlay Hints don't work
 
 Currently, only inlay hints provided by TypeScript are supported. TypeScript inlay hints are disabled by default and needs to be enabled using the settings under the `typescript.inlayHints` namespace, for example, to enable inlay hints for parameter names, you would do the following:


Princesseuh:
Hmm, I don't see why this is necessary. On most setups, you won't have any other formatters enabled. The Astro extension will by default be used. What was your setup trying to use to format??

Khaledgarbaya:
I installed the extension but didn't get auto formatting, for example on save, unless I add that to my settings.
I use prettier in vs code for formatting.

Princesseuh:
> I installed the extension but didn't get auto formatting, for example on save, unless I add that to my settings. I use prettier in vs code for formatting.

Sure, but what extension was VS Code trying to use to format? VS Code uses the only formatter available if there's only one. Prettier does not advertise itself as being Astro compatible by default unless you configure it to do so, in which case it seems obvious to me that you need to make sure you're using the formatter you want (and VS Code will prompt you on format)

Khaledgarbaya:
It was not very obvious for me and others. Prettier is wildly used by folks out there so I thought maybe it gets a special mention in the extension docs otherwise users will think that the extension is broken but I might be wrong.

raae:
From the list below I felt like it should work out of the box, but it didn't and I got help from @Khaledgarbaya on Twitter via his answer to another person: https://x.com/PaulieScanlon/status/1704207301819662457?s=20
CleanShot 2023-09-20 at 21 48 33@2x

Princesseuh:
Okay, I don't think I've made myself clear on what I'm asking 😅

What this configuration does is set the default formatter for Astro files to be the Astro extension. 

However: 
- If there's only one formatter for a file format, VS Code will use that formatter by default.
- No other extensions on the marketplace is compatible with Astro files, by default (Prettier's VS Code extension *can be*, if you specifically configure it to be)
- If for some reason, you have multiple formatters installed, but none are set as default, VS Code will prompt you to choose.

What this configuration does is simply reiterate the default behaviour of VS Code. As such, I don't understand *why* it would fix anything, nor why it would be necessary to have

Khaledgarbaya:
Sounds Good, maybe something is wrong in our vs code configuration ¯\_(ツ)_/¯. Closing the PR, and thank you for your time

PR #654, opened 2023-09-24
fix(parser): nested `<script>`/`<style>` tags are not extracted

F0rce:
## Changes

The `language-server` only embedded `<script>`&`<style>` tags on root level. This PR adds recursion on extracting/embedding all `<script>`&`<style>` tags from the `htmlDocument`. Using `.children` recursively.

## Testing

Change was tested by nesting multiple `<div>`'s and then trying to autocomplete in an `<script>` or `<style>` tag. I didn't find any relevant tests for that sort of case, if wanted I could add one.

## Docs

bug fix only

### DIFF
--- a/packages/language-server/src/core/parseJS.ts
+++ b/packages/language-server/src/core/parseJS.ts
@@ -65,12 +65,15 @@ function extractEmbeddedJSFilesRecursively(
 	fileName: string,
 	snapshot: ts.IScriptSnapshot,
 	roots: Node[],
-	array: VirtualFile['embeddedFiles'] = []
+	array: VirtualFile['embeddedFiles'] = [],
+	level: string[] = []
 ): VirtualFile['embeddedFiles'] {
 	for (const [index, root] of roots.entries()) {
-		if (root.tag !== 'script') continue; // Early return if tag is not 'script'
+		const currentLevel = [...level, index.toString()]; // Append current index to the level
+		const newFileName = `${fileName}.${currentLevel.join('.')}`;
 
 		if (
+			root.tag === 'script' &&
 			root.startTagEnd !== undefined &&
 			root.endTagStart !== undefined &&
 			isIsolatedScriptTag(root)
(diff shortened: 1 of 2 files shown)


packages/language-server/src/core/parseJS.ts
```diff
@@ -17,44 +17,11 @@ export function extractScriptTags(
...
-		if (
-			root.tag === 'script' &&
-			root.startTagEnd !== undefined &&
-			root.endTagStart !== undefined &&
-			isIsolatedScriptTag(root)
-		) {
-			const scriptText = snapshot.getText(root.startTagEnd, root.endTagStart);
-
-			embeddedJSFiles.push({
-				fileName: fileName + `.${index}.mts`,
```
F0rce, line 31: I am not quite sure if this is the right approach. Should there be some sort of level visible in the filename? E.g. `Header.astro.0.mts` (root level), but if the root level would have some children `Header.astro.0.0.mts` (first child).

packages/language-server/src/core/parseJS.ts
```diff
@@ -94,6 +61,57 @@ export function extractScriptTags(
...
+
+		if (
+			root.startTagEnd !== undefined &&
+			root.endTagStart !== undefined &&
+			isIsolatedScriptTag(root)
+		) {
+			const scriptText = snapshot.getText(root.startTagEnd, root.endTagStart);
+
+			array.push({
+				fileName: fileName + `.${index}.mts`,
```
F0rce, line 81: > Deleted original comment, because I placed it wrong.

I am not quite sure if this is the right approach. Should there be some sort of level visible in the filename? E.g. `Header.astro.0.mts` (root level), but if the root level would have some children `Header.astro.0.0.mts` (first child).

Princesseuh: The names do need to be unique, but the structure doesn't really matter, feel free to achieve this in any way you see fit!

Princesseuh:
Thank you! This looks great. I just went ahead and did some refactoring and added some tests here: https://github.com/withastro/language-tools/pull/656. I couldn't do it in your PR since you fork doesn't have access for the original maintainers, but nonetheless the commits have proper credits so you'll get the proper credits on your profile. Thank you very much!

PR #671, opened 2023-10-24
Update compiler

Princesseuh:
## Changes

> Blocked until more users are on version at least 2.10.8 of Astro

What the title says!

## Testing

Tests should pass!

## Docs

N/A


delucis:
Not sure if this still has the limitation of #629, but just in case, providing the data that in the last 30 days about 18% of users were on versions lower than 2.10.8

ematipico (approved):

Princesseuh:
> Not sure if this still has the limitation of #629, but just in case, providing the data that in the last 30 days about 18% of users were on versions lower than 2.10.8

It still does, thank you for posting this! I wasn't intending to release this right now, just was updating the old PR.

PR #679, opened 2023-10-30
feat(ts-plugin): compatible with TS 5.3

johnsoncodehk:
## Changes

Changes compatible with TS 5.3:

- No more hack `getScriptKind` (Fixed by https://github.com/microsoft/TypeScript/pull/55716)
- Refresh external files based on the new updateLevel parameter to identify newly created files (Fixed https://github.com/vuejs/language-tools/issues/3555)


Princesseuh (approved):
Looks great, thank you!

PR #681, opened 2023-11-01
fix: formatting messing up inline scripts

Princesseuh:
## Changes

Formatting was wrongly enabled inside inline script tags capabilities, we don't want formatting on virtual files because Prettier handles the entire `.astro` file itself

## Testing

Added a test

## Docs

N/A

### DIFF
--- a/packages/language-server/test/units/parseJS.test.ts
+++ b/packages/language-server/test/units/parseJS.test.ts
@@ -37,7 +37,7 @@ describe('parseJS - Can find all the scripts in an Astro file', () => {
 		expect(scriptTags.length).to.equal(0);
 	});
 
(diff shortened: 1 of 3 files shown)


packages/language-server/test/units/parseJS.test.ts
```diff
@@ -36,4 +36,27 @@ describe('parseJS - Can find all the scripts in an Astro file', () => {
 
 		expect(scriptTags.length).to.equal(0);
 	});
+
+	it('returns the proper capabilities for inline script tags zzzz', () => {
```
natemoo-re, line 40:
```suggestion
	it('returns the proper capabilities for inline script tags', () => {
```

natemoo-re (approved):
LGTM!

Issue #644, opened 2023-09-01
🐛 BUG: LSP couldn't locate the Astro in a monorepo


Stradi:
### Describe the Bug

Hi.

I am using Astro 3.0.5 in a monorepo-ish directory structure. Why monorepo-ish, because I'm not using any external tool to manage monorepo. I'm just creating folders for each app and package.

```
package.json -> Just a metadata information about the project itself.
apps/
  |- frontend/
      |- node_modules/
      |- package.json
      |- astro.config.mjs
      |- ...
  |- backend/
      |- node_modules/
      |- package.json
      |- ...
packages/
  |- some-helper-package
      |- node_modules/
      |- package.json
      |- ...
```

The problem is that the Astro page in `apps/frontend/src/pages/index.astro` has all the globals (`Astro` for example) recognized but `apps/frontend/src/pages/auth/sign-up.astro` page doesn't. After I move the Astro project (frontend directory) in `packages/` instead of `apps/` the error goes away.

VSCode shows a warning like this:

```Couldn't find Astro in workspace "/home/batin/projects/project-name". Experience might be degraded. For the best experience, please make sure Astro is installed into your project and restart the language server.```

The output panel for Astro Language Server is also showing an error:

```/home/batin/projects/project-name seems to be an Astro project, but we couldn't find Astro or Astro is not installed```

I've dig through the source code of Astro Language Server and found that language server only checks `packages/` directory for monorepos. I've also tested this by renaming `apps/` into `packages/` and it works. Here is the related code:

[packages/language-server/src/utils.ts#L27-L44](https://github.com/withastro/language-tools/blob/main/packages/language-server/src/utils.ts#L27-L44)

```ts
// If we couldn't find it inside the workspace's node_modules, it might means we're in the monorepo
try {
    astroPath = getPackagePath('./packages/astro', basePaths);

    if (!astroPath) {
        throw Error;
    }

    version = require(path.resolve(astroPath, 'package.json')).version;
} catch (e) {
    // If we still couldn't find it, it probably just doesn't exist
    console.error(
        `${basePaths[0]} seems to be an Astro project, but we couldn't find Astro or Astro is not installed`
    );

    return undefined;
}
```

I think LSP should check all directories except some big ones (like `node_modules`) and try to find Astro project that way.

### Steps to Reproduce

1. Create `apps/` directory in root folder
2. `cd` into `apps/` and run `pnpm create astro` with default options
[...]

PR #682, opened 2023-11-03
fix: improve Astro detection to avoid false warnings

Princesseuh:
## Changes

Improve the Astro detection to only attempt to find Astro in relevant contexts and better handle incomplete packages

Fix https://github.com/withastro/language-tools/issues/704
Fix https://github.com/withastro/language-tools/issues/644 (as far as I could tell? I tried reproducing it in both npm and pnpm and couldn't, everything worked perfectly fine.)

## Testing

Tested manually, we can't really test this in the monorepo because `astro` is a devDep of some things and so resolution isn't really accurate compared to a normal Astro project.

## Docs

N/A

### DIFF
--- a/.changeset/rich-adults-sparkle.md
+++ b/.changeset/rich-adults-sparkle.md
@@ -1,0 +1,6 @@
+---
+'@astrojs/language-server': patch
+'astro-vscode': patch
+---
+
+Improve detection of Astro in complex monorepos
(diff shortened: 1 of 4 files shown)


alexanderniebuhr (approved):
As far as I can follow the code, this looks fine. /lgtm

Is there a way to test this before releasing the extension?

natemoo-re (approved):
LGTM!

PR #686, opened 2023-11-12
Fix ignored minimal severity flag

Igloczek:
## Changes

When you use `astro check --minimumSeverity error`, it's supposed to filter logs based on the severity level, but it's showing all logs, regardless of this setting. 

I've tweaked it so that the default is now `hint`, which keeps things as they were. But when you set `--minimumSeverity error`, it'll only show the logs that are errors, just like you'd expect.

## Testing

There aren't any automated tests for this feature yet, so I just tested this manually.

## Docs

This update is more about fixing the behavior to match what we say in the `--help` output, and I didn't find any specific documentation that needs updating for this.

### DIFF
--- a/.changeset/popular-shirts-fix.md
+++ b/.changeset/popular-shirts-fix.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/language-server": patch
+---
+
+Fix ignored minimal severity flag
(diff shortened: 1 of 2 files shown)


Princesseuh (approved):
Looks great, thank you!

PR #701, opened 2023-11-16
feat: upgrade to Volar 2.0 alpha

johnsoncodehk:
## Changes

- What does this change?


spacedawwwg:
@johnsoncodehk @Princesseuh This update to Volar 2.0 alpha, `astro check` throws up loads of type issues that aren't being reported in the editor?

setting my overrides to `"@astrojs/language-server": "2.5.3"` has allowed us to get by for now - but my question is: Why would we suddenly start throwing up so many issues that weren't there before?

Princesseuh:
> @johnsoncodehk @Princesseuh This update to Volar 2.0 alpha, `astro check` throws up loads of type issues that aren't being reported in the editor?
> 
> setting my overrides to `"@astrojs/language-server": "2.5.3"` has allowed us to get by for now - but my question is: Why would we suddenly start throwing up so many issues that weren't there before?

This PR hasn't been released. You're probably experiencing the following issue: https://github.com/withastro/language-tools/issues/727

Issue #587, opened 2023-06-18
🐛 BUG: `onclick` JavaScript statements are "linked"?


TheOtterlord:
### Describe the Bug

Creating 2 HTML elements with `onclick` events can cause the first one to affect parsing of the second in VSCode. In this example, the second `onclick` attribute displays `';' expected.ts(1005)` as an error highlighting `document`.

```
<button onclick="document.querySelector('#modal').style.display='block'">Open modal</button>
<button onclick="document.querySelector('#modal').style.display='block'">Open modal</button>
```

### Steps to Reproduce

1. `npm create astro@latest -- my-app --template minimal -y`
2. Paste this code in `src/pages/index.astro` or any Astro file
```astro
<button onclick="document.querySelector('#modal').style.display='block'">Open modal</button>
<button onclick="document.querySelector('#modal').style.display='block'">Open modal</button>
```
3. `';' expected.ts(1005)` will appear as an error on line 2

torchsmith:
Same issue here:

```
<div
	class="open"
	onclick="toggleMenu()"
>
.... bunch of html
<div
	class="open"
	onclick="toggleMenu()"
>
... bunch of html
```

Princesseuh:
Hmm, I definitely get why this happens (All the inline scripts are technically in one file that get type checked, since they operate in the same scope), to which a simple partial solution is to append a virtual semi-colons at the end of every context, however that would still cause some errors to spill over, I wonder how VS Code avoids this in their HTML support. Will have to investigate, funny one

PR #724, opened 2023-12-15
fix(parseJS): Suffix inline attributes with a semicolon to avoid errors leaking

Princesseuh:
## Changes

What the title says! This is not perfect, but then again, it's not perfect in VS Code's HTML support either:

image

If this is enough for VS Code, it's enough for me. At least, for now.

Fix https://github.com/withastro/language-tools/issues/587

## Testing

Tested manually

## Docs

N/A


ematipico (approved):
I'm happy too!

Issue #727, opened 2023-12-16
🐛 BUG: astro check reports incorrect errors on HTMLAttributes


DerYeger:
### Describe the Bug

While this worked as expected in Astro 4.0.4, `astro check` now reports entirely valid JSX attributes as invalid HTMLAttributes.
I've created a tiny reproduction here: https://stackblitz.com/edit/github-ckecs3?file=package.json.
Running `npm run check` will show that both `onClick` and `className` are invalid, even though both are in fact valid JSX props.

### Steps to Reproduce

1. Run `npm create astro@latest -- --template framework-preact`
2. Add `"check": "astro check:"` to the `scripts` section
3. Run `npm run check`
4. The valid `onClick` attributes are considered errors. Now, changing one of the `class` attributes to `className` (which is valid) will also show that as an error.

Princesseuh:
Hmm, seems like it's not taking into account the `tsconfig.json` correctly, for some reason. To be clear, this also didn't work in the past, it's just that I recently fixed a bug where it ignored errors completely inside certain files.

millette:
I'm seeing the same error with my solidjs components.

Princesseuh:
I'll take a look once I'm back to work next week (if someone doesn't before me). Apologies for the inconvenience. In the meantime, pinning `@astrojs/check` to `0.3.2` would at least make it not fail (but would go back to the behavior where it didn't check certain files)

flmn:
I have the same problem with React component.

jlarmstrongiv:
Unfortunately, downgrading to `0.3.2` didn’t fix it for me

Princesseuh:
> Unfortunately, downgrading to `0.3.2` didn’t fix it for me

You might need to override the language-server version manually, unfortunately. Hoping to have a fix out today, sorry for the delay.

craigjennings11:
@jlarmstrongiv I had to revert my `pnpm-lock.yaml` file as well as downgrade to get it working again. Downgrading alone left some other dependency upgraded that made it keep failing.

Princesseuh:
Okay, this is proving to be much harder than anticipated due to bugs in TypeScript itself. It'll take more time to figure out, so in the meantime I'll revert the `astro check` fix to what it did before (where it skipped over .js(x) files). 

Sorry for the inconvenience, TypeScript really wasn't made to have multiple JSX typings in the same project.

knpwrs:
[...]

Issue #715, opened 2023-12-05
🐛 BUG: Auto add missing imports dont work well


TheElegantCoding:
### Describe the Bug

in my vscode config i have this line of code, this let me save the file and let vscode auto import the missings imports, but this dont work well

```json
{
  "editor.codeActionsOnSave": {
    "source.fixAll": true,
    "source.addMissingImports": true
  },
}
```

as you can see in the photo, this add the import in the top of the line not where it should go

![image](https://github.com/withastro/language-tools/assets/70770072/16364189-c1cc-4402-b23e-5b2df66d891a)

### Steps to Reproduce

```json
{
  "editor.codeActionsOnSave": {
    "source.fixAll": true,
    "source.addMissingImports": true
  },
}
```

Princesseuh:
Thank you for this issue! Mapping imports is always a bit tricky, ha.

PR #733, opened 2023-12-20
Properly account for new TSX prefix

Princesseuh:
## Changes

Fixing `astro check`'s checking behaviour so it checks `.js(x)` files properly revealed a problem: We force `astroHTML`'s types in every file types. This is not a problem in the editor really, because in Astro projects you kinda don't rely on the types of JSX components inside Astro files, but in `astro check`, this is obviously wrong.

In https://github.com/withastro/compiler/pull/917, I made it so our TSX is now prefixed with a JSX pragma, that way we don't need to force a JSX setting on every file, our files just set the proper runtime types to use automatically. Unfortunately, this is quite rigid and requires exporting the JSX types from somewhere named `package/jsx-runtime`, so I did that in Astro's JSX runtime in https://github.com/withastro/astro/pull/9501

This is all great and awesome, however a lot of our mapping around auto imports and code actions depended on the frontmatter being the first thing in the generated TSX and JSX pragmas needs to be on the first lines of TSX files, so in https://github.com/withastro/compiler/pull/931, I made it so the compiler now returns the ranges of the TSX inside the virtual file, that way we can map directly to there. This was a fair amount of work.

The diff is larger than it might seems due to many dependencies updates needed for kinda unrelated things, Volar's current essentially pre-1.0 state suggest updating everything to be at the same exact version when you update one thing. The Astro version used for tests was also updated to account for the jsx-runtime export

Fix https://github.com/withastro/language-tools/issues/727 (I'll remove the `astro check` hotfix workaround in a separate PR)
Fix https://github.com/withastro/language-tools/issues/715 by accident too

## Testing

Tests should pass! Updated the unit tests for the utils to match the new ranges they return (since they now return ranges in the TSX vs ranges in the Astro file, they might look inaccurate, but as far as I can tell, everything is okay!)

## Docs

N/A


packages/language-server/src/core/astro2tsx.ts
```diff
@@ -109,7 +109,7 @@ function getVirtualFileTSX(
 										semantic: true,
 										navigation: true,
 										structure: true,
-										format: true,
+										format: false,
```
Princesseuh, line 112: Unrelated fix, but we don't want TSX formatting since we handle it all in Prettier

packages/language-server/src/core/index.ts
```diff
@@ -61,9 +61,9 @@ export function getLanguageModule(
 							module: ts.ModuleKind.ESNext ?? 99,
 							target: ts.ScriptTarget.ESNext ?? 99,
 							jsx: ts.JsxEmit.Preserve ?? 1,
```
Princesseuh, line 63: Keeping this one for now because I don't see a way to set it through a pragma. Since this is only for type checking, it should be fine to keep it at Preserve, I think

packages/language-server/test/fixture/enhancedDiagnostics.astro
```diff
@@ -0,0 +1,6 @@
+---
```
Princesseuh, line 1: This needs to be in a real file now because otherwise it cannot seems to resolve `astro/jsx-runtime`, not sure why. Might be something with how fake files works in tests.

natemoo-re (approved):
As much as I am able to wrap my head around this it LGTM!

.changeset/good-ants-switch.md
```diff
@@ -0,0 +1,7 @@
+---
+'@astrojs/language-server': minor
+'@astrojs/ts-plugin': minor
+'astro-vscode': minor
+---
+
+Internally type Astro files individually instead of applying a specific configuration to every file loaded in the Astro language server, this should generally result in more accurate types when using JSX frameworks. But should generally be an invisible change for most users.
```
natemoo-re, line 7: Slightly more concise changeset
```suggestion
Enables more accurate types when using JSX-based frameworks. This internal refactor to Astro's JSX types will be an invisible change for most users, but fixes a number of type checking problems for users of other JSX frameworks.
```

packages/language-server/src/core/index.ts
```diff
@@ -85,7 +117,7 @@ export class AstroFile implements VirtualFile {
 	fileName: string;
 	mappings!: VirtualFile['mappings'];
 	embeddedFiles!: VirtualFile['embeddedFiles'];
-	astroMeta!: ParseResult & { frontmatter: FrontmatterStatus };
+	astroMeta!: AstroMetadata;
```
MoustaphaDev, line 120: TIL `!:` is a thing 😄

MoustaphaDev, line 120: TIL `!:` is a thing 😄

MoustaphaDev (approved):
TIL `!:` is a thing 😄

PR #799, opened 2024-02-17
map compiler location to LSP range

martrapp:
## Changes

For compiler warnings, the visual markers were also displayed on some characters in the following line. Maybe I misunderstood something here, but I fixed it to the best of my knowledge :)
 
## Testing

Manually tested.
There are test for Compiler Points but those functions seem to be outdated not to be in use anywhere?

## Docs

n.a., bug fix

### DIFF
--- a/.changeset/smart-cycles-join.md
+++ b/.changeset/smart-cycles-join.md
@@ -1,5 +1,7 @@
 ---
 "@astrojs/language-server": patch
+"@astrojs/check": patch
+"astro-vscode": patch
 ---
 
 Fixes mapping from compiler location to LSP range.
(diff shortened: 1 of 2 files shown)


Princesseuh:
> There are test for Compiler Points but those functions seem to be outdated not to be in use anywhere?

Those are for the AST positions, we used them not too long ago for mapping code actions, but they're currently unused. I'm keeping them around because it is quite likely that we'll use them in the future

.changeset/smart-cycles-join.md
```diff
@@ -0,0 +1,5 @@
+---
+"@astrojs/language-server": patch
```
Princesseuh, line 2:
```suggestion
"@astrojs/language-server": patch
"@astrojs/check": patch
"astro-vscode": patch
```

Princesseuh (approved):
Looks great, thank you!

PR #804, opened 2024-02-22
Makes astro check --tsconfig understand relative file names

martrapp:
## Changes

`astro check` has an `--tsconfig` option that takes a file parameter.
Now it also understands relative file names.

## Testing

Manually tested

## Docs

bugfix / n.a.


Princesseuh (approved):
Awesome. looks great!

PR #818, opened 2024-03-08
chore: update biome

ematipico:
## Changes

- Updated `biome` to the latest.
- `biome` now formats `package.json` files. Happy to revert the changes. They are different from prettier because prettier just format them with `JSON.strigify(content, null, 2)`.
- I removed the eslint prettier packages because they aren't needed anymore.
- Biome now handles Astro and Svelte files, with some limitations. For example the indentation in the svelte file is different (and expected). Happy to revert it if not wanted.

## Testing

The CI should still pass

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -1,7 +1,7 @@
 module.exports = {
 	parser: '@typescript-eslint/parser',
-	extends: ['plugin:@typescript-eslint/recommended', 'prettier'],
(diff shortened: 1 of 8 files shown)


Princesseuh:
I would prefer if we ignore fixtures, because some of the tests depend on precise positions that formatting could prevent from working

ematipico:
@Princesseuh done!

PR #826, opened 2024-03-13
fix `tmLanguage`: text.astro pattern

goulvenclech:
## Changes

In this case : 
```jsx
blabla<div>word</div>
{true && <div>word</div>}blabla
```

Both `blabla` have the scope `source.astro` instead of `text.astro`, which is the scope in HTML. Fixed the `text.astro` pattern to match that.

## Testing

Snap test -> `packages/vscode/test/grammar/fixtures/text.astro.snap`

## Docs

N/A, this PR will match the documented behaviour.

### DIFF
--- a/.changeset/wet-beers-change.md
+++ b/.changeset/wet-beers-change.md
@@ -1,0 +1,5 @@
+---
+"astro-vscode": patch
+---
+
+text.astro pattern matches text at the beginning and end of the document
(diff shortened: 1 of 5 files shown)


goulvenclech:
@Princesseuh shouted at me, so I added a test :'(

Princesseuh (approved):
Looks great, thank you!

Issue #708, opened 2023-11-22
🐛 BUG: Highlighting without closing tag


cdtut:
### Describe the Bug

Braces aren't highlighted properly without closing tag.

### Steps to Reproduce

These braces aren't highlighted properly because there is no closing tag. There are html tags like `link` that don't have closing tags.

```
{check && <tag>}
```

These braces highlighted properly.

```
{check && <tag />}
{check && <tag></tag>}
```

swift502:
Not just braces, this completely breaks style and script block highlighting.
![comp](https://github.com/withastro/language-tools/assets/24359130/890b80e3-1f04-481c-972a-7957014e8d49)

PR #828, opened 2024-03-13
AstroJSX syntax [WIP]

goulvenclech:
## Changes

Astro expression syntax is similar to JSX, but it's not JSX. So, we need to create a new syntax for it, including the JSX syntax as a base, and adding some Astro-specific patterns over it.

This will allow us to fix some issues much more easily, such as #708 . Also, the syntax will match the actual behaviour of Astro expressions, like basic HTML comments - not supported by JSX in expressions, but Astro does.

## Testing

Will be tested. I swear.

## Docs

N/A, this PR will make the syntax match the actual behaviour of Astro expressions. But the code is commented.

### DIFF
--- a/packages/vscode/package.json
+++ b/packages/vscode/package.json
@@ -162,8 +162,7 @@
           "source.stylus": "stylus",
           "source.js": "javascript",
           "source.ts": "typescript",
-          "source.json": "json",
-          "source.tsx": "typescriptreact"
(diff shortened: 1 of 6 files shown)


Princesseuh:
This repo has been merged into the main Astro monorepo, if you'd still like for this PR to happen, please re-submit it there.

Issue #713, opened 2023-11-30
🐛 BUG: Can't toggle comments in braces


cdtut:
### Describe the Bug

When there is code like this and you toggle comment on the middle line it shows like the 2nd example instead of 3rd. This is not valid so have to manually add and remove comments on both sides which is slow and terrible experience or delete the lines which you can forget to add again.

```
{array.map(each => <>
  <div>{each}</div>
</>)}

{array.map(each => <>
  // <div>{each}</div>
</>)}

{array.map(each => <>
  {/* <div>{each}</div> */}
</>)}
```

### Steps to Reproduce

Toggle comment on middle line.

Princesseuh:
Unfortunately, we don't control how the comment toggling works inside expression because we use VS Code's TSX syntax configuration there, which controls the commenting behavior.

I think it is possible to fix it somehow, but it's quite challenging.

cdtut:
Maybe can allow `//` to be a comment?

Or if it can be fixed because it's happening many times a day. Very frustrating development experience to lose something so basic and common like automatic commenting shortcut. Something we should be able take for granted.

Many issues with extension are because of JSX syntax. Desperate need for alternative syntax that does not rely on strange JSX issues maybe use html as basis for syntax. You think is good idea too?

Princesseuh:
> Maybe can allow `//` to be a comment?
> 
> Or if it can be fixed because it's happening many times a day. Very frustrating development experience to lose something so basic and common like automatic commenting shortcut. Something we should be able take for granted.
> 
> Many issues with extension are because of JSX syntax. Desperate need for alternative syntax that does not rely on strange JSX issues maybe use html as basis for syntax. You think is good idea too?

There would be just twice as many problems, especially in the type checking behavior with another syntax. Syntax files are just very complicated and fairly rigid.

cdtut:
If alternative syntax is too complicated and so is this one then I don't know. Easy commenting and syntax highlighting and completion are core to development experience. Syntax complication means it's not easy to use plain text editor it's better to use extension but building something so complicated seem like it's not realistic to have smooth development experience with astro.

khanakia:
any update on this it's quite furstrating

Princesseuh:
> any update on this it's quite furstrating

I would share updates if they were any. If you are interested in contributing, I would be happy to guide you through what's needed.

khanakia:
I would be happy to help but I have no idea how to fix this.


[...]

PR #858, opened 2024-04-23
fix(syntax): Apply JSX latest rules for our language

Princesseuh:
## Changes

Add a few more things from TSX's syntax to our rules that are not inherited due to a VS Code bug. 

Fixes https://github.com/withastro/language-tools/issues/713 and a few unreported issues (ex: `>>>=` wasn't properly highlighted

## Testing

Tested manually, you can't test this inside grammar tests because you don't see the underlying language being applied, just the `source.tsx` scope.

## Docs

N/A

### DIFF
--- a/.changeset/forty-flowers-drum.md
+++ b/.changeset/forty-flowers-drum.md
@@ -1,0 +1,5 @@
+---
+"astro-vscode": patch
+---
+
+Fixes commenting shortcut not using the proper comments inside expressions in certain cases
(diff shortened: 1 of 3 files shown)


Princesseuh:
Test failure is unrelated to the change (the test in question doesn't test this), not sure what's happening

Princesseuh:
Oh, it's a Node change https://github.com/nodejs/node/issues/52554

matthewp (approved):

PR #863, opened 2024-05-02
feat: Upgrade to Volar 2.2

johnsoncodehk:
## Changes

Upgrade to Volar 2.2 💪


packages/vscode/package.json
```diff
@@ -49,10 +49,6 @@
       }
     ],
     "commands": [
-      {
-        "command": "astro.reloadProjects",
-        "title": "Astro: Reload Projects"
-      },
```
Princesseuh, line 55: Does this command not exist anymore in 2.2? People use it fairly often to workaround caching issues

johnsoncodehk: Yes it was removed in the refactor, I didn't expect it to be used, I will add it back.

johnsoncodehk: Added by [`18a302a` (#863)](https://github.com/withastro/language-tools/pull/863/commits/18a302adb862a400903e7329ddbb7ae7ea75c078)!

Princesseuh:
Tried this PR locally and it seems to work, however I noticed an issue with the Prettier formatting, seems like it's not loading the Prettier config anymore.

Not quite sure what's up with the tests, but since the features do work locally I can only assume this is a bug in the test utils / tests

Princesseuh:
@johnsoncodehk My last commit fixes some of the tests. However, it seems like there's an issue with `openInMemoryDocument` in Volar 2.2, all the remaining tests that don't pass uses it.

Strangely enough, if I change the fake URI to be one of a file that does exist, no matter its content, then it works.

Princesseuh (approved):
Awesome work, this looks great!

Issue #869, opened 2024-05-12
🐛 BUG: Moving files results in formatting issues in Svelte files


danielo515:
### Describe the Bug

When I move a svelte file that contains some typescript definitions, like an interface for example, many times the code gets messed, and if you undo, it will not go back to it's original state.
This seems to be related to a bug in refactoring imports, because it happens when the language server tries to correct the imports.

### Steps to Reproduce

1. `npm init astro`
2. add svelte integration
3. create a svelte file with imports
4. Move the file to a nested folder or to a parent folder
5. The code gets broken

This happens in VSCode by the way, not sure about other usages of the language server

Princesseuh:
This extension does not run inside Svelte files. Is it the code inside the Svelte files that get messed up?

danielo515:
Yes, it also breaks autocompletion in svelte files. The extension may not
directly deal with svelte files, but if I disable it then the problem goes
away

---
https://danielorodriguez.com

Princesseuh:
Does it break auto completions or does it break what you described in the first post?

danielo515:
It breaks automatic imports and refactors. Suggestions do work (but very
bad)

---
https://danielorodriguez.com

Princesseuh:
It seems to be some sort of conflict between the Svelte TypeScript plugin and the Astro one. It seems like both plugins are trying to do the refactor at the same time and so it ends up generating broken code. For what it's worth, you can actually undo the changes if you undo the move, and not the latest change. The LSPs are doing it correctly, as far as I can tell, just the TypeScript plugins.

I cannot reproduce a difference regarding auto imports / completions with or without the Astro plugin, so not sure on that one. Since auto imports in the respective files would be handled by the LSPs, I'm not sure how the Astro extension could affect the Svelte one in that regard.

Princesseuh:
This should be fixed in 2.9.1. Apologies for the inconvenience. If you have the time, I would appreciate if you updated your review on the VS Code marketplace to reflect this.

danielo515:
That is awesome! Thank you very much. I will update it, sorry about it. I wrote it in a moment where I was very frustrated

Princesseuh:
No worries, it happens! We're all humans here

PR #871, opened 2024-05-14
fix: don't generate mappings for Svelte and Vue files

Princesseuh:
## Changes

We wrongfully had mappings for Vue and Svelte files, so when we updated files, we could sometimes try to update Svelte and Vue files, but really, we don't know what to do there, so we shouldn't.

Fixes https://github.com/withastro/language-tools/issues/869

## Testing

Tested manually, this is quite tough to test in the testing suite because we can't quite mock what the client does

## Docs

N/A


bholmesdev (approved):
Happy to not do things!

PR #878, opened 2024-05-28
feat: highlight Astro code blocks in MDX

remcohaszing:
## Changes

This adds support for syntax highlighting `astro` tagged code blocks inside MDX files.

````mdx
This is MDX

```astro
---
const frontmatter = 'This is Astro frontmatter'
---

<div>This is an Astro template</div>
```
````

image

## Testing

With the [MDX extension](https://marketplace.visualstudio.com/items?itemName=unifiedjs.vscode-mdx) installed, open the Astro language tools repo and create an Astro code block inside an MDX file.

## Docs

I don’t know what documentation would be needed for this.

### DIFF
--- a/.changeset/warm-lemons-search.md
+++ b/.changeset/warm-lemons-search.md
@@ -1,0 +1,5 @@
+---
+"astro-vscode": patch
+---
+
+Add syntax highlighting for `astro` tagged code block inside MDX files.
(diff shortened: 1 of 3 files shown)


Princesseuh:
I know we do it that way in Markdown, but I don't know why we redefine the frontmatter and stuff. Couldn't we just apply `source.astro` to code blocks with the astro language?

remcohaszing:
Without it the frontmatter content isn’t highlighted. I don’t understand why.

I did simplify it a bit now.

PR #881, opened 2024-06-14
Fix/restore a few caching tests

machty:
As part of my work on Volar-izing Glint I discovered that `didChangeWatchedFiles` needs to be present as a server capability in order for `didChangeWatchedFiles` to work properly within specs. I noticed Astro pending some of these specs, so I took a quick stab to see whether adding this capability would fix things. Indeed, it fixed a few test cases, but there are two that are still broken (I don't have time to dig in myself at this time but if I have any insights after working on Glint a bit more I'll share them here).


Princesseuh:
Fixed one of the test that was not passing, the other one is a genuine issue in Volar / typescript-auto-import-cache cc @johnsoncodehk

PR #883, opened 2024-06-19
Add `@astrojs/check` README

delucis:
## Changes

- Adds a README to the `@astrojs/check` package similar to the ones we use for Astro integrations (inspired by someone who found the package on npm and didn’t know what it was)
- Adds `description` and `homepage` fields to `package.json` too

## Testing

n/a metadata change only

## Docs

It’s all docs I guess.

### DIFF
--- a/packages/astro-check/README.md
+++ b/packages/astro-check/README.md
@@ -1,6 +1,6 @@
 # @astrojs/check ✅
 
-Type checking and diagnostics for your Astro project.
+Type checking and diagnostics for Astro projects.
 
 ## Documentation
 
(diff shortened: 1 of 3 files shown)


florian-lefebvre (approved):

packages/astro-check/README.md
```diff
@@ -0,0 +1,36 @@
+# @astrojs/check ✅
+
+Type checking and diagnostics for Astro projects.
```
Princesseuh, line 3: It might be useful to specify that this powers the `astro check` command, otherwise people might think it's an actual type checker

delucis: Something like this? I’ll admit I followed the other READMEs that aim to give zero docs, just links.

```suggestion
Provides the `astro check` command for running type checking and diagnostics in Astro projects.
```

PR #893, opened 2024-07-08
[bugfix] Executing `pnpm test` on workspace folder with a space on the name doesn't work `ENOENT: no such file or directory`

alexrecuenco:
## Changes

- My workspace folder had a space `External Projecs` and executing `pnpm test`, and with shell:true, arguments of `child_process.spawn` are not escaped.
- I believe that shell: true might be unnecessary. If it isn't, another solution would have to be to do manual escaping of each command argument...

```bash

packages/vscode
astro-vscode:test: > node scripts/build-grammar.mjs
astro-vscode:test: 
astro-vscode:test: [09:54 AM] ✔ updated /Workspace/External Projects/language-tools/packages/vscode/syntaxes/astro.tmLanguage.json
astro-vscode:test: node:fs:581
astro-vscode:test:   return binding.open(
astro-vscode:test:                  ^
astro-vscode:test: 
astro-vscode:test: Error: ENOENT: no such file or directory, open '/Workspace/External'

```

## Testing

If the tests don't run successfully on the test matrix on all systems, this is clearly wrong.

## Docs

bug fix only on a test script

### DIFF
--- a/packages/vscode/test/grammar/test.mjs
+++ b/packages/vscode/test/grammar/test.mjs
@@ -17,6 +17,7 @@ const allGrammars = [...grammars, ...dummyGrammars];
 
 /**
  * @param  {Parameters<typeof spawn>} arg
+ * @returns {Promise<number | null>}
  */
 function promisifySpawn(...arg) {
 	const childProcess = spawn(...arg);
@@ -45,10 +46,9 @@ async function snapShotTest() {
 
 	const code = await promisifySpawn(process.platform === 'win32' ? 'pnpm.cmd' : 'pnpm', args, {
 		stdio: 'inherit',
-		shell: true,
 	});
 
-	if (code > 0) {
+	if (code && code > 0) {
 		process.exit(code);
 	}
 }


Princesseuh:
`shell: true` is necessary because of https://github.com/nodejs/node/issues/52554

alexrecuenco:
I see.  I imagined it had to be something to do with windows, but I would have never guessed that issue! In that case, I'll do the space quoting when I have some time.

Should I keep shell:true if the platform is not windows? It might be unnecessary. 

 https://nodejs.org/api/child_process.html#spawning-bat-and-cmd-files-on-windows

Princesseuh:
That seems fine to me!

alexrecuenco:
Hi @Princesseuh , I haven't had much time. I would need some help to test this, since I don't have a windows machine. I think what I did is sane? at the very least better (assuming I followed the instructions correctly). 

If it is too complex, we could just disable the escape in non-windows and leave that as a TODO for when someone with a windows machine encounters that issue and has a bit of time to fix it

This is unrelated, I think? but when I run `pnpm test` all the `Content Intellisense - ` tests fail, they all receive a `[]`. Is there something I should be doing on my setup to get those tests to work?

Princesseuh:
Yeah, there's a failure on main on a content intellisense test, haven't had the time to investigate yet, but it's definitely unrelated.

Princesseuh (approved):
Looks great, thank you! Apologies for the delay.

Issue #862, opened 2024-04-26
🐛 BUG: autocomplete does not work with inline style when Japanese text is included in an element.


tsukashu:
### Describe the Bug

When Japanese text is included in an element, autocompletion with inline style attributes does not work **after** the location where the Japanese text appears.And, it may not work on any element after that.

### Steps to Reproduce

1. Install VSCode Version: 1.88.x (I use Portable version)
2. `npm create astro@latest "--" --template minimal -n`
3. Add Latest Astro VSCode Extension and install npm packages include the template.
4. Add some Japanese elements.
5. First Japanese elements, autocomplete with inline style works fine, but after all elements will not work.

https://github.com/withastro/language-tools/assets/80503344/45cef9a8-d79e-49b1-bf4d-acea20c1f2d1

thank you for your efforts.

Princesseuh:
Might be related to https://github.com/withastro/language-tools/issues/660, thank you for reporting this issue, we'll take a look as soon as possible!

PR #897, opened 2024-07-09
fix(js/css): Fixes intellisense inside inline styles and scripts

Princesseuh:
## Changes

We calculate inline scripts and styles (inside HTML events and `style` attributes respectively) using the positions from the AST the compiler returns. This is fine, however the offset it returns consider multiple bytes characters, unlike LSP (well, it's in reality more complicated than that because of encoding and stuff, but it's still somewhat that)

This PR changes it so we uses the Point returned instead, which do consider lines and columns character by character

Fixes https://github.com/withastro/language-tools/issues/862

## Testing

Added a bunch of tests

## Docs

N/A

### DIFF
--- a/.changeset/weak-kiwis-rule.md
+++ b/.changeset/weak-kiwis-rule.md
@@ -1,0 +1,7 @@
+---
+"@astrojs/language-server": patch
+"@astrojs/check": patch
+"astro-vscode": patch
+---
+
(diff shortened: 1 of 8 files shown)


delucis (approved):
I believe in you! 🌻

Princesseuh:
Superceded by https://github.com/withastro/language-tools/pull/900, added the tests from this PR to it

Issue #800, opened 2024-02-18
Adding Scripts Affects TypeScript Intellisense on `.astro` file


iamhectorsosa:
### Astro Info

```block
Astro                    v4.4.0
Node                     v20.11.0
System                   macOS (arm64)
Package Manager          pnpm
Output                   static
Adapter                  none
Integrations             @astrojs/tailwind
                         @astrojs/react
```

### Describe the Bug

Adding Scripts Affects TypeScript Intellisense on `.astro` file. The snippet below has been included in Layout.astro or index.astro files. In both situations with or without directives I've noticed that the TypeScript intellisense is completely disabled. Missing imports are not detected, no intellisense on any props for the imported components, etc.

```html
<script is:inline>
  const getThemePreference = () => {
    if (typeof localStorage !== "undefined" && localStorage.getItem("theme")) {
      return localStorage.getItem("theme");
    }
    return window.matchMedia("(prefers-color-scheme: dark)").matches
      ? "dark"
      : "light";
  };
  const isDark = getThemePreference() === "dark";
  document.documentElement.classList[isDark ? "add" : "remove"]("dark");

  if (typeof localStorage !== "undefined") {
    const observer = new MutationObserver(() => {
      const isDark = document.documentElement.classList.contains("dark");
      localStorage.setItem("theme", isDark ? "dark" : "light");
    });
    observer.observe(document.documentElement, {
      attributes: true,
      attributeFilter: ["class"],
    });
  }
</script>
```

### What's the expected result?

TypeScript is not affected by the inclusion of script tags in `.astro` files

### Link to Minimal Reproducible Example

I don't have a GitHub repo at the time with this.

### Participation

- [ ] I am willing to submit a pull request for this issue.

Princesseuh:
I'm not quite sure what the exact cause is, but you can workaround the issue by putting the script tag below the content.

iamhectorsosa:
Thanks for looking into it. Yeah, also importing works for this. Still a strange issue.

PR #900, opened 2024-07-11
fix: scripts affecting intellisense

Princesseuh:
## Changes

Requires https://github.com/withastro/compiler/pull/1019

This PR reworks how script and style are extracted from files, using the compiler's information instead of crawling through various ASTs. This makes file updates simpler, much faster, and also more correct as the ASTs sometimes had wrong or misleading positions due to how they were being generated.

Fixes https://github.com/withastro/language-tools/issues/800
Fixes https://github.com/withastro/language-tools/issues/862
This PR also fixes an issue where CSS selection ranges didn't work inside style tags, reported on Discord

## Testing

Tests should pass!

## Docs

N/A

### DIFF
--- a/packages/language-server/src/core/parseCSS.ts
+++ b/packages/language-server/src/core/parseCSS.ts
@@ -1,169 +1,45 @@
-import type { ParentNode, ParseResult } from '@astrojs/compiler/types';
-import { is } from '@astrojs/compiler/utils';
+import type { TSXExtractedStyle } from '@astrojs/compiler/types';
 import type { CodeInformation, VirtualCode } from '@volar/language-core';
 import { Segment, toString } from 'muggle-string';
-import type ts from 'typescript';
-import type { HTMLDocument, Node } from 'vscode-html-languageservice';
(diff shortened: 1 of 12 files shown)


packages/language-server/src/core/parseCSS.ts
```diff
@@ -1,169 +1,45 @@
...
-	return embeddedCSSCodes;
+export function extractStylesheets(styles: TSXExtractedStyle[]): VirtualCode {
+	return mergeCSSContexts(styles);
 }
 
-/**
- * Find all inline styles using the Astro AST
- * Inline styles are styles that are defined in the `style` attribute of an element.
- * TODO: Merge this with `findEmbeddedCSS`? Unlike scripts, there's no scoping being done here, so merging all of it in
- * the same virtual file is possible, though it might make mapping more tricky.
```
Princesseuh, line 114: This is the case now in the new format, which improves intellisense for class names and stuff when you have multiple style tags in the same file. Kinda unlikely, but still funny!

matthewp (approved):

PR #911, opened 2024-07-18
feat: put emmet in JSX mode to avoid expressions completions

Princesseuh:
## Changes

By forcing JSX mode on Emmet instead of HTML, we don't get the useless completions inside expressions. I'm not sure 100% what else it changes, as it's not documented, but reading through Emmet's source code I couldn't find anything too bad.

Close https://github.com/withastro/language-tools/discussions/835

## Testing

For some reason, I can't seem to be able to get emmet completions in the tests. I remember something similar happening in the previous language-server as well, I'm not too sure why that happens!

## Docs

N/A

### DIFF
--- a/.changeset/ten-adults-sparkle.md
+++ b/.changeset/ten-adults-sparkle.md
@@ -1,0 +1,6 @@
+---
+"@astrojs/language-server": patch
+"astro-vscode": patch
+---
+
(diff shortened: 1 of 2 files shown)


delucis (approved):
`js`t e`x`ceptional ✅

ematipico (approved):

Issue #916, opened 2024-07-25
🐛 BUG: Latest Astro VS Code extension update causing linting errors for SCSS in <style lang="scss"> tags


LushawnDev:
### Describe the Bug

I thought there was something I changed in my project, but after some elimination I realised it was the Astro extension. I was getting linting errors like 

- "Unknown at rule @usecss(unknownAtRules)" 
- "at-rule or selector expectedcss(css-ruleorselectorexpected)"

in all of my style tags using SCSS when using ` <style lang="scss">`. I thought I updated a package or something that caused an error as the issue was not page or component specific, as it is happened in every Astro file where I was using scss.

After I realised it was coming from my VS Code, I saw the Astro extension was updated yesterday. I downgraded to the previous version and the error no longer exists.

### Steps to Reproduce

1. `npm init astro` using template <NAME> and install the latest Astro extension in VS Code
2. Create a page using `.astro` file extension
3. Create a  <style lang="scss"> and use some SCSS in it
5. Error! You should see linter warnings around any SCSS used as it seems to be expecting CSS only still despite the attribute.

Princesseuh:
After further investigation, it turns out that the problem here is only that we show some errors that we used to not show. In a way, that's great, but it also unveiled that our SCSS support actually didn't work.

I'll fix both the errors showing and see to improve the SCSS (and LESS, since it's handled similarly) support. Apologies for the inconvenience.

Princesseuh:
The errors are now fixed in the latest version of the extension!

PR #923, opened 2024-07-29
Adds support for SCSS and LESS intellisense inside style tags

Princesseuh:
## Changes

> Require https://github.com/withastro/compiler/pull/1032

A long time ago (before the port to Volar), this was actually supported, but we lost it without noticing. This PR adds it back.

Fixes https://github.com/withastro/language-tools/issues/916

## Testing

Added tests

## Docs

N/A


delucis (approved):
More is LESS 🚀

matthewp (approved):

Issue #920, opened 2024-07-26
🐛 BUG: `astro check` does not support TSC project references


codethief:
### Describe the Bug

This is a follow-up to my message [here](https://github.com/withastro/language-tools/issues/674#issuecomment-2250936856). 

Passing a "solution-style" tsconfig.json to `astro check`, which [`references` other projects (tsconfigs)](https://www.typescriptlang.org/docs/handbook/project-references.html), will make the type checker silently accept type errors associated with types imported from those other projects.

The issue might be related to https://github.com/withastro/language-tools/issues/649 since project references [rely on emitting type declarations](https://github.com/microsoft/TypeScript/issues/49844#issuecomment-2249921055).

### Steps to Reproduce

(This is a placeholder ticket for now; I will update it soon with example code to reproduce the issue.)

codethief:
@Princesseuh I added a minimal reproduction [here](https://github.com/codethief/tsconfig-playground/tree/9cdaa547fa59978c590adb5bf08cedeee9f281ba/2024-07-27-astro-with-project-references).

PR #938, opened 2024-08-18
feat: Update to Volar 2.4.0

johnsoncodehk:
## Changes

- What does this change?


packages/language-server/src/check.ts
```diff
@@ -145,7 +146,13 @@ export class AstroCheck {
 		const services = [...createTypeScriptServices(this.ts), createAstroService(this.ts)];
 
 		if (tsconfigPath) {
-			this.linter = kit.createTypeScriptChecker(languagePlugins, services, tsconfigPath);
+			const includeProjectReference = false; // #920
```
johnsoncodehk, line 149: @Princesseuh To be consistent with old behavior, a false value is used here, but fixing #920 requires changing to true. Since this will impact performance, you may want to pass a flag to decide whether to enable it.

eidam:
Ola, I believe this might also fix https://github.com/withastro/language-tools/issues/939 ?

PR #944, opened 2024-08-26
chore(astro-check): remove unused fast-glob dep

bluwy:
## Changes

I don't think it matters in practice since `fast-glob` is also used by astro and language-server, but this helps keep the exact deps used concise.

## Testing

existing tests should pass

## Docs

n/a


Princesseuh:
The `fast-glob` usage in the language-server is wrong anyway, it'll have to get removed eventually. Thank you!

PR #953, opened 2024-09-12
refactor(language-server): unused resolveLanguageServiceHost hook

johnsoncodehk:
## Changes

resolveLanguageServiceHost hook is a bad abstraction. Its problem is that the hook logic cannot be universal to TS plugin. We plan to deprecate resolveLanguageServiceHost hook in Volar 2.5 and migrate to a more reasonable setup hook in this PR. setup hooks are also available for TS plugins, but PR did not implement this in order not to change the old behavior.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,19 @@
   },
   "packageManager": "pnpm@8.6.2",
   "pnpm": {
+    "overrides": {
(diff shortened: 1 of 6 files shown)


Princesseuh (approved):
Looks great, I like this API!

PR #954, opened 2024-09-20
chore(language-server): move `@volar/typescript` as dev dep

bluwy:
## Changes

Lame PR 😅  but `@volar/typescript` was only used as types so we move it as a dev dep. It's still used transitively by `@volar/language-server` and `@volar/kit`, so not removed from the dependency graph

## Testing

Manually checked that it's not used in the compiled js or dts files.

## Docs

n/a


Princesseuh (approved):
I'll allow it

Issue #946, opened 2024-08-26
VSCode formatting doesn't work right after running `npm astro-create@latest`


mihailthebuilder:
### Astro Info

```block
Astro                    v4.14.5
Node                     v20.9.0
System                   Windows (x64)
Package Manager          npm
Output                   static
Adapter                  none
Integrations             none
```

### Describe the Bug

When I create a new Astro project with `npm create-astro@latest`, formatting in VSCode doesn't work. After I downgrade to an earlier version (`4.11.3`), it starts working again. I then upgrade to the latest version, and it works.

### What's the expected result?

VSCode formatting works.

### Link to Minimal Reproducible Example

https://github.com/mihailthebuilder/astro-format-bug

### Participation

- [ ] I am willing to submit a pull request for this issue.

Princesseuh:
This happens because of `yaml-language-server`, used now with content intellisense, having a dependency on Prettier 2, which cause formatting to fail.

I'm not sure what we'll be able to do for this, might need to fork upstream 😔

bholmesdev:
Unfortunately kabirsync's fix didn't work for me. I intalled the [community prettier plugin](https://github.com/withastro/prettier-plugin-astro) to manually run `prettier . --write` from the CLI. @Princesseuh do you know of any workarounds in VS Code to enforce prettier on save rather than the Astro language tools? I haven't been able to figure this out.

PR #960, opened 2024-10-01
fix: fallback to bundled Prettier when finding Prettier 2

Princesseuh:
## Changes

`yaml-language-server` uses Prettier 2.x to power its formatting, this confuse our formatting because we expect Prettier 3. Really, they shouldn't be using Prettier in the first place (see https://github.com/redhat-developer/yaml-language-server/issues/933), but alas, they do. We don't use their formatting at all, but still this affects us. 

This PR refactor the package loading to not try to load Prettier 2 when it's there, which should fix both this situation and also when people have dependencies using Prettier 2. Installing Prettier in your own repo also still works, as `require.resolve` will find it first before more nested versions.

Fix https://github.com/withastro/language-tools/issues/946

## Testing

Tests should still pass, and tested locally that it outputs the error message correctly

## Docs

N/A

### DIFF
--- a/.changeset/grumpy-planes-sip.md
+++ b/.changeset/grumpy-planes-sip.md
@@ -1,0 +1,7 @@
+---
+"@astrojs/language-server": patch
+"@astrojs/check": patch
+"astro-vscode": patch
+---
+
+Fixes formatting not working by default in certain circumstances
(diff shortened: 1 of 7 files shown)


bluwy (approved):
Looks great!

PR #962, opened 2024-10-04
feat: distribute the language server and related packages as ESM

Princesseuh:
## Changes

VS Code famously does not support ESM inside extensions, however, this does not prevent us from publishing everything related as ESM, as other editors can support it just fine.

This PR migrates everything apart from two packages:

- The VS Code extension (duh)
- The TypeScript plugin
  - I couldn't figure it out, if I wrote it as ESM, but converted it to CJS with esbuild TSServer complained that the plugin was wrongly shaped. Looking around the ecosystem, everyone seems to write them as CJS so, oh well.
  
This PR also updates other packages, I had a weird problem with something not being able to understand the syntax, so just went ahead and updated every devDeps.

## Testing

I tried it manually, it's very possible that as publish time everything break though.

## Docs

N/A

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,13 +18,13 @@
     "lint": "eslint ."
   },
   "devDependencies": {
-    "@biomejs/biome": "1.8.1",
-    "@changesets/cli": "^2.26.1",
-    "prettier": "^3.2.5",
-    "turbo": "1.10.2",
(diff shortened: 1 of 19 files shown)


Princesseuh:
This is proving to be way too complicated, will wait a bit more...

PR #1023, opened 2025-09-24
Add logic for auto-updating schema config from collection metadata

delucis:
## Changes

This PR adds some logic to try and simplify use of Astro-generated JSON schemas for data files in content collections. We [document](https://docs.astro.build/en/guides/content-collections/#collection-json-schemas) these schemas, but the VS Code config to use them isn’t so lovely.

Sharing mainly for feedback at this point as there are several less than optimal things here, and I may be doing things wrong in any case.

Notes:

- AFAICT there is no way to set the `json.schemas`/`yaml.schemas` settings in an ephemeral “per session” way. We can only update a user’s workspace config directly (i.e. the one that lives in `.vscode/settings.json` in a project). That’s pretty annoying as it means the generated changes this added logic creates would show up in e.g. git diffs. All the worse because…

- Currently I’m using the `.astro/collections/collections.json` metadata file we generate for the [“content intellisense” experimental feature](https://docs.astro.build/en/reference/experimental-flags/content-intellisense/). But this doesn’t give me e.g. globs for which files are impacted. This makes the auto-updated settings **_verbose_**, because we list all impacted files, and will mean every new file incurs a change to `settings.json` to commit. Here’s an example of enabling this in the Astro Docs repo which uses YAML files for i18n:

	```json
	  "yaml.schemas": {
	    "./.astro/collections/i18n.schema.json": [
	      "/src/content/i18n/ar.yml",
	      "/src/content/i18n/de.yml",
	      "/src/content/i18n/es.yml",
	      "/src/content/i18n/en.yml",
	      "/src/content/i18n/hi.yml",
	      "/src/content/i18n/fr.yml",
	      "/src/content/i18n/it.yml",
	      "/src/content/i18n/ko.yml",
	      "/src/content/i18n/ja.yml",
	      "/src/content/i18n/pl.yml",
	      "/src/content/i18n/ru.yml",
	      "/src/content/i18n/zh-cn.yml",
	      "/src/content/i18n/zh-tw.yml",
	      "/src/content/i18n/pt-br.yml",
	      "__astro__"
	    ]
	  }
	```
	
	If configuring this manually you’d usually do this instead, which is much cleaner:
	
	```json
	"yaml.schemas": {
	  "./.astro/collections/i18n.schema.json": "/src/content/i18n/*.yml"
	}
	```

- Because we write to a user’s config, we need a way to distinguish auto-generated and user-authored config to avoid overwriting a user’s settings. There’s no really good way to do this. I’ve hacked around it here by adding an `__astro__` boolean property for JSON schema entries, and adding `__astro__` as the last entry to each array for YAML schemas. Pretty gross.

Anyway, mainly sharing so this isn’t lost on my hard drive. If anyone has ideas of how to make this less rubbish, feel free to share!

## Testing

Manually tested in VS Code.

## Docs

TO DO

### DIFF
--- a/packages/vscode/src/client.ts
+++ b/packages/vscode/src/client.ts
@@ -1,4 +1,5 @@
 import * as path from 'node:path';
+import { fileURLToPath } from 'node:url';
 import * as protocol from '@volar/language-server/protocol';
 import type { LabsInfo } from '@volar/vscode';
 import {
@@ -14,6 +15,7 @@ import * as vscode from 'vscode';
 import * as lsp from 'vscode-languageclient/node';
 
 let client: lsp.BaseLanguageClient;
+let collectionsWatcher: vscode.FileSystemWatcher;
 
 type InitOptions = {
 	typescript: {
@@ -73,6 +75,10 @@ export async function activate(context: vscode.ExtensionContext): Promise<LabsIn
 		contentIntellisense: hasContentIntellisense,
 	} satisfies InitOptions;
 
+	if (hasContentIntellisense && rootPath) {
+		await initializeCollectionSchemaAutoconfig(rootPath);
+	}
+
 	const clientOptions = {
 		documentSelector: [
 			{ language: 'astro' },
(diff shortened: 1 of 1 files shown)


delucis:
@Princesseuh Do you have any feedback on this general approach? As it is, it’s kind of garbage, and I don’t really know how to make it less garbage. Otherwise I’ll let it die for now.

Princesseuh:
> @Princesseuh Do you have any feedback on this general approach? As it is, it’s kind of garbage, and I don’t really know how to make it less garbage. Otherwise I’ll let it die for now.

Hmm, I'll admit that I don't know either. I figured that there would be a way to set it per session, but if there's not, well...

PR #1033, opened 2025-10-15
feat: migrate tests to node:test

Princesseuh:
## Changes

This migrates our two main packages (language-server and astro-check) to use node:test for tests, like the main repo. Unfortunately, the two other packages with tests depend on running VS Code itself using its test library and while I did get it working with node:test, it stalled on Windows and required a lot of flaky code to even make it work. So for now they still use Mocha.

## Testing

It's all tests!

## Docs

N/A


packages/language-server/test/script.ts
```diff
@@ -0,0 +1,95 @@
+import fs from 'node:fs/promises';
```
Princesseuh: Is it intended at this time that there's two of this, they're slightly different and are both fork of the script we have in core. When I merge the repo inside the monorepo, I'll update both packages to use the "official" version

